PR #59, opened 2020-04-09
ENG-375 - Low level dev workflow for generated widgets

rimselis:


docs/draft/microfrontends/change-watcher.adoc
```diff
@@ -0,0 +1,73 @@
+= Tutorial: Blueprint generated widgets' change watcher overview
+:toc:
+
+For an easier development you can use a small script to watch for changes inside the
```
gidesan, line 4: inside the... ?

docs/draft/microfrontends/change-watcher.adoc
```diff
@@ -0,0 +1,73 @@
...
+```
+  "dependencies": {
+    "node-notifier": "^6.0.0",
+    "shelljs": "^0.8.3"
+  }
+```
+
+`node-notifier` is just for ease of usage.
+
+Change watcher itself expects the arguments passed via cli command (see above). Change watcher tracks `path/to/entando-blueprint/generators/entity-server` folder. When there are any chhnges, the `jhipster entity ENTITY [--regenerate] --force` command is called and the widgets are rebuilt.
```
gidesan, line 25:
```suggestion
Change watcher itself expects the arguments passed via cli command (see above). Change watcher tracks `path/to/entando-blueprint/generators/entity-server` folder. When there are any changes, the `jhipster entity ENTITY [--regenerate] --force` command is called and the widgets are rebuilt.
```

PR #73, opened 2020-05-13
Integrate entando labs front page into vuepress

raxkaynan:
* Integrates the entando labs front page into vuepress
* Changes base docs URL to "/docs/" as root ("/") would now be the entando labs front page


joewhite101:
@raxkaynan are you sure this should be here? Won't the front page live at a different URL?

raxkaynan:
> @raxkaynan are you sure this should be here? Won't the front page live at a different URL?

@joewhite101 Talked to Ed about it and he said it's the front page for docs.entando.com.

PR #98, opened 2020-06-15
Eng 558 container limits configurable

firegloves:
- added generic "Minimum required resources" section at the beginning of the quickstart
- added detailed "Define resource limits" section illustrating how to update resources requests and limits
- updated predefined entando.yml file used to create the quickstart app

### DIFF
--- a/vuepress/docs/docs/getting-started/README.md
+++ b/vuepress/docs/docs/getting-started/README.md
@@ -186,6 +186,15 @@ IP=$(hostname -I | awk '{print $1}')
 sed -i "s/192.168.64.25/$IP/" entando.yaml
 ```
 
+### Define resource limits
+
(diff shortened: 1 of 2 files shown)


vuepress/docs/docs/getting-started/README.md
```diff
@@ -180,6 +186,33 @@ IP=$(hostname -I | awk '{print $1}')
 sed -i "s/192.168.64.25/$IP/" entando.yaml
 ```
 
+### Define resource limits
```
nshaw, line 189: Could we move this to a dedicated page, perhaps under docs/reference and then linked from here? I think this is great information but probably not needed on the getting started page which we're trying to keep as short as possible.

nshaw: @firegloves , I can try to fold this into my 6.2 work if that would be helpful but let me know what you think.

nshaw:
Closing in favor of https://github.com/entando/entando-docs/pull/110

PR #99, opened 2020-06-15
ENG-352 how to customize bundle title, thumbnail and description

Kerruba:
adds a tutorial on how to change bundle title, thumbnail or description from the custom-resource


nshaw:
Hey, @Kerruba. Is this 6.2-specific or does this instruction work for 6.1 as well? Based on the 6.1 CRD I have installed it looks like this works for 6.1.

Kerruba:
Hi @nshaw, yeah this is valid for 6.1

Kerruba:
I think should be mergable, let me know if you need anymore feedback

PR #106, opened 2020-06-18
Endoc 25 past prs

nshaw:
Folding in some of the 6.1 changes we lost during the docs restructure.


vuepress/docs/tutorials/ecr/tutorials/ecr-deploy-use-plugin-and-mfe-without-bundle.md
```diff
@@ -0,0 +1,170 @@
...
+> The bundle population with the micro frontends requires some time to
+> be processed. You should be able to follow the progress of the
+> operation on screen.
+
+## 3. Generate a docker image for your microservice
+
+JHipster uses the JIB Maven plugin to generate a docker image for your
+microservice.
+
+By default the output image will use the name of your application and
```
Kerruba, line 59: Now during the blueprint wizard we ask the user to provide a custom organization (if they want) or use the default one which is `entando`. Do you want to mention that here somehow?

vuepress/docs/tutorials/ecr/tutorials/ecr-deploy-use-plugin-and-mfe-without-bundle.md
```diff
@@ -0,0 +1,170 @@
...
+the `0.0.1-SNAPSHOT` tag, eg. `entando/jhipster@0.0.1-SNAPSHOT`.
+
+    ./mvnw -Pprod clean package jib:dockerBuild [-Djib.to.image=<your-image-name]
+
+> **Note**
+>
+> By default, the organization used for the image is `entando` but you
```
Kerruba, line 66: I think we can mention here that in the new wizard, user is asked for an `organization` where to publish the docker image. That organization is set by default to `entando` but the user can override it quite easily. Using their custom organization solves issues like "I'm not authorized to publish a docker image in the `entando` organization on dockerhub", or "I've built my custom image but the microservice I'm deploying is still referencing the `entando` organization"

PR #119, opened 2020-06-29
ENDOC-7 add versions

nshaw:
This may be easier to review here: https://es-entando.github.io/. 

There were a few minor content changes but the major steps included:
1) Add versioning per a POC Ed put together. See the vuepress/README for a few notes on that.
2) Make all of the links under v6.1 relative so the content can easily be copied to create a new version.
3) Copy v6.1 to v6.2
4) Fold in the work-in-progress from entando-v6.2-wip into the v6.2 directories.

The look-and-feel could be improved (e.g. Versions menu) but I consider this MVP. 

There is one annoyance with some top-level links, e.g. https://es-entando.github.io/docs, which temporarily show a 404 before resolving. I spent several hours hunting for a fix but the workarounds other users have applied for this (due to SSR mismatch) didn't fix it here. I opened ENDOC-57 in case we have energy to pursue that later.


vuepress/README.md
```diff
@@ -52,3 +57,23 @@ As an example, we'll be adding 2 related documentation files, named `README` and
...
+### Versions
+We have a simple versioning system in place currently. Creating a new version involves the following steps:
+1. Copy the directories under the current version (e.g. v6.2) to a new version.
+2. Modify `.vuepress/config.js` 
+   1. Add the new version to the menu. A `Beta` suffix is recommended until the new version is live.
+   2. Add sidebars for the new version.
+3. At launch time:
+   1. Modify the 3 manual redirect pages under `docs`, `docs/getting-started` and `tutorials` to point to the now active version
+   2. Remove `Beta` from the menu in config.js
+   3.   Update `SpecialLayout.vue` to pint to the new active version of `Docs` and `Tutorials`
```
es-entando, line 70: *point*

nshaw: Fix pushed.

walterambu:
Great! I also suggest we include 5.3 documentation (it needs to be visible). From what I understood from Alessandro Pinna, it should be a copy of 5.2 documentation. Maybe we can have a clone of 5.2 documentation?

nshaw:
@joewhite101 , I like the latest/next idea. I put a summary of our conversation in slack to see if anyone has an opinion but I can make that change later today.

@w-caffiero-entando, I can easily add a link to the old versions (https://es-entando.github.io/old-version/old-version.html#_entando_5_2_0) but it's a bit misleading. There is older 6.1 and 6.0 content there which should really be removed and we'll need to resurrect the asciidocs source to create a 5.3 copy. I've played around with the asciidocs generation but I'm seeing quite a few differences from what is generated now and what is being served by vuepress. I'll need others to indicate if that effort is worth it at this point so I'll raise this on the arch call today for a quick opinion. Some of the changes are obvious based on the git history (docs.entando.com renamed to dev.entando.org, favicon updated, etc.) but others are not (all the anchors use underscores instead of dashes in the regenerated version, etc.).

nshaw:
> Looks good overall. I think we should make the most recent one "latest" or some other invariant text that will always show the latest value rather than "6.2 (Beta)" so that there is a persistent link to the latest and greatest.
> 
> Also would recommend moving Versions to be either the left entry or the far right entry. Not easy to find in current position.

@joewhite101 , I switched v6.2 to next as we discussed. The Versions select is moved to the left as well. https://es-entando.github.io/next/docs/

nshaw:
@w-caffiero-entando , I reworked the older asciidocs a bit and added 5.3 to the menu.  https://es-entando.github.io/next/docs/

nshaw:
Discussed with @walterambu and good to go as well.

PR #120, opened 2020-06-30
Endoc 49 add gke

joewhite101:
Add baseline install docs for GKE


joewhite101:
Need to rebuild from 6.2 baseline

PR #133, opened 2020-07-10
Endoc 39 communication

es-entando:
Updated micro frontend communication and angular to react tutorial


joewhite101:
Is this deployed somewhere? Pretty big change. We'll need to test.

Also, was widget communication removed from 6.1 on purpose?

es-entando:
@joewhite101 it can be deployed to my GitHub site if you and Nathan are done with the changes that are staged there.

@gidesan this is the first tutorial where I've made a substantial change to your writeup so I've added you as a reviewer. In this case, I merged your Angular to React Communication tutorial into the Communicate Between Micro Frontends tutorial. I thought the Angular to React Communication tutorial made more sense as an add-on compared to duplicating all of the steps required to create a React subscriber and then add it as a widget to a page in the App Builder.

I was pretty split on this decision because I'd prefer not to diverge from the original content too much, and I do like them as separate tutorials. However, making the Angular to React Communication tutorial a self-contained tutorial would require duplication of > 50% of the content in Communication Between Micro Frontends so I felt that merging them was the better choice.

nshaw:
We're running a training this week so let's hold on merging/publishing this until at least Friday to avoid impacting the current team.

es-entando:
Once the micro frontend tutorials are in a good place, I will check all of those files to ensure I didn't "lose" any files in the reorganization. I can remove those duplicate files at that time. Sounds good on waiting until after the training is over to merge.

nshaw:
I should have mentioned this is visible on staging now: https://es-entando.github.io/next/tutorials/micro-frontends/communication.html#publisher

PR #169, opened 2020-08-06
ENDOC-79 EKS edits

nshaw:


nshaw:
@joewhite101 , this is a work in progress. My IAM steps were pretty different to get things working so we may need to discuss. I'm waiting for my domain to propagate but everything after step 10 looks like standard Entando at that point.

nshaw:
This is deployed to https://es-entando.github.io/next/tutorials/devops/installation/elastic-kubernetes-service/eks-install.html

PR #176, opened 2020-08-12
Endoc 92 add image zoom

nshaw:


nshaw:
@joewhite101 , this is deployed to staging, e.g. on https://es-entando.github.io/v6.2/docs/ you can click on the arch image and it expands to fit the available window. We have a couple of screenshots that are so long it's not perfect but this should be an 80%+ solution.

PR #178, opened 2020-08-14
Endoc 75 add release notes

nshaw:


nshaw:
This is visible on staging - https://es-entando.github.io/v6.2/docs/releases

PR #180, opened 2020-08-26
ENDOC-96 update azure nginx instructions

joewhite101:
Nginx deprecated the steps that we have in azure. This updates to the new repos and naming from here:
https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx

Which is where the deprecation message points.

### DIFF
--- a/vuepress/docs/next/tutorials/devops/installation/azure-kubernetes-service/azure-install.md
+++ b/vuepress/docs/next/tutorials/devops/installation/azure-kubernetes-service/azure-install.md
@@ -125,7 +125,7 @@ curl -sfL https://github.com/entando-k8s/entando-helm-quickstart/archive/v6.2.0.
 helm template my-aks-app --namespace=entando ./ > my-aks-app.yaml
 ```
 5. Deploy Entando via `kubectl create -f my-aks-app.yaml`
-6. Watch Entando startup `kubectl get pods -n entando --watch`
+6. Watch Entando startup `watch kubectl get pods -n entando`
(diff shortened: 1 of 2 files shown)


vuepress/docs/next/tutorials/devops/installation/azure-kubernetes-service/azure-install.md
```diff
@@ -123,7 +125,7 @@ curl -sfL https://github.com/entando-k8s/entando-helm-quickstart/archive/v6.2.0.
 helm template my-aks-app --namespace=entando ./ > my-aks-app.yaml
 ```
 5. Deploy Entando via `kubectl create -f my-aks-app.yaml`
-6. Watch Entando startup `kubectl get pods -n entando --watch`
+6. Watch Entando startup `watch kubectl get pods -n entando`
```
nshaw, line 128: @joewhite101 What was the thinking on this change? We have a couple dozen cases of kubectl --watch in the docs and in my testing watch isn't as flexible since it doesn't respect aliases.

joewhite101: I will roll it back. I find the output of that structure a lot better but if it breaks things we can keep it as is.

joewhite101:
> Probably want to update the 6.2 docs to match. If you have rsync handy you can run yarn docs:sync62 to make that a little easier, notes in the top vuepress/README.

Done

PR #193, opened 2020-09-11
Eng 1252 update plugin descriptor

Kerruba:


vuepress/docs/next/docs/ecr/ecr-bundle-details.md
```diff
@@ -74,43 +74,21 @@ following structure.
 
 ## Plugin Descriptor
 
-A plugin should be described using the [EntandoPlugin CustomResource
```
nshaw, line 77: Why did you remove the reference to the CRD definition?  It seems like that would still be useful.

Kerruba: Because the descriptor had changed and is not anymore like the Entando Plugin custom resource. 

Maybe we can move that link later in the documentation when I explain how the PluginDescriptor is converted to an EntandoPlugin 

What you think?

nshaw: That works for me. I just liked pointing to the descriptor where it makes sense.

Kerruba: I've added a small reference in another piece of the documentation, let me know if that's ok with you :+1:

nshaw (approved):

PR #194, opened 2020-09-16
ENDOC-103 Fix some language in getting started. Add theme plugin code…

nshaw:
…-copy. Improve language in entando-apis.md.


nshaw:
Unfortunately the code-copy is a theme plugin and doesn't apply to the landing page because of the way it's built as a non-themed straight-Vue custom template. I don't think that's a deal breaker since we should be switching the simplest option over to the entando cli before long but the code-copy will be broadly useful for detailed instructions, tutorials, etc.

It's visible in staging now: https://es-entando.github.io/v6.2/docs/getting-started

joewhite101 (approved):
LGTM

PR #208, opened 2020-10-12
ENDOC-66 Switch staging site

nshaw:
This will switch our staging site from https://es-entando.github.io/ to https://entando.github.io/entando-docs/.


es-entando (approved):
Looks good.

PR #211, opened 2020-10-14
ENG-1370 retrocompatible installation old plugin descriptor + ENG-1380 deploymentBaseName

firegloves:


w-caffiero-entando (changes requested):
@firegloves @nshaw I propose to not document the old version at all.

It's deprecated (users shouldn't use it) and it's currently supported only for compatibility with  descriptors contained in already existing bundles.

nshaw:
> @firegloves @nshaw I propose to not document the old version at all.
> 
> It's deprecated (users shouldn't use it) and it's currently supported only for compatibility with descriptors contained in already existing bundles.

That makes sense to me unless there's a compelling reason to document it for current clients. If that's the case I'd suggest we add this just to the 6.2 version of the docs and mark it deprecated. I suspect we'll be cutting the 6.3 version of the docs soon and we could leave it out there.  @firegloves , @w-caffiero-entando , thoughts?

firegloves:
@nshaw @w-caffiero-entando may be an option to place a line like this?
"the previous plugin descriptor version is now deprecated. It's still supported but we encourage you to migrate to the newer version"

nshaw:
There's a product question whether we actually plan to support it. I'd favor documenting it for 6.2 as Deprecated but not include it in next/6.3. I can make that change if @w-caffiero-entando is okay with the approach.

w-caffiero-entando:
> I'd favor documenting it for 6.2 as Deprecated but not include it in next/6.3. I can make that change if @w-caffiero-entando is okay with the approach.

Uhm, on 6.2 can't be deprecated as it's the only version supported.

I'm fine keeping it on 6.3 but deprecated, but I'd prefer to keep on the main page only the new one and plus link to the deprecated version.

nshaw:
Thanks for clarifying. I misunderstood that part. We should probably call that out in the Release Notes then since it could be a breaking change in 6.4.

w-caffiero-entando:
> Thanks for clarifying. I misunderstood that part. We should probably call that out in the Release Notes then since it could be a breaking change in 6.4.

Agree

nshaw:
@w-caffiero-entando , @firegloves I updated this with the deprecation warnings. Visible at https://entando.github.io/entando-docs/next/docs/ecr/ecr-bundle-details.html#plugin-descriptor and https://entando.github.io/entando-docs/v6.2/docs/ecr/ecr-bundle-details.html#plugin-descriptor.

nshaw (approved):

firegloves:
I added some docs for ENG-1380

w-caffiero-entando (approved):

PR #236, opened 2020-11-25
Upgrade vuepress and dependencies

nshaw:
Bumping up the specific dependencies in package.json (manually) and implicit dependencies via yarn upgrade. Will address one specific dependabot warning but we're also a bit overdue. I did some light testing and everything looks okay, the plugins are working fine.


joewhite101 (approved):
LGTM

PR #249, opened 2020-12-17
ENDOC-19 Add full text search

nshaw:
This enables full text search on the site. It's enabled in staging https://entando.github.io/entando-docs/ and all of the content on the pages is now indexed whereas before it was just page and section titles. Phrase searches also work much better now.

### DIFF
--- a/vuepress/docs/.vuepress/config.js
+++ b/vuepress/docs/.vuepress/config.js
@@ -36,6 +36,10 @@ module.exports = {
         align: 'top',
         selector: 'div[class*="language-"] pre'
     },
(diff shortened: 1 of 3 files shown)


nshaw:
@w-caffiero-entando , I took a look at the Google alternative but I think this is more Vuepress-y and let's us keep control of the content. It might be better for SEO too although I'm not sure on that. Flexsearch has a ton of options so there a lot of ways we could tune the search experience (like maybe by excluding old versions, or boosting new ones, etc.). I played with enabling contextual search but didn't see a major difference and didn't want to put a lot of time into it.

es-entando:
Verified on staging that it's displaying a lot more results. Nice update.

es-entando (approved):

PR #252, opened 2021-01-04
Endoc 175 add hubspot tracking

nshaw:


nshaw:
@joewhite101 , this is ready for review but I'll hold on merging until I hear about the cookiebot account and whether we can actually use the same account for dev.entando.org.

joewhite101 (approved):
LGTM

nshaw:
Changed my mind, moving ahead with this change. Worst case I'll disable the cookie banner before launching 6.3 but I think we need both of these changes and this will unblock another PR I'm working.

PR #257, opened 2021-01-08
Endoc 182 minor edits

nshaw:
Two tweaks here: 1) Add the code-copy button to the landing pages. Bit of forked code there since the landing pages don't use the theme so they can't use theme plugins. 2) Replace a couple more usages of wget in favor of curl.

### DIFF
--- a/vuepress/deploy-staging.sh
+++ b/vuepress/deploy-staging.sh
@@ -18,7 +18,6 @@ npm run docs:build
 git restore docs/.vuepress/config.js
 git restore docs/.vuepress/components/css/main.css
 git restore docs/.vuepress/components/SpecialLayout.vue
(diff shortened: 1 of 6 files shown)


nshaw:
This is visible on staging now, e.g. https://entando.github.io/entando-docs/openshift.html or https://entando.github.io/entando-docs/

joewhite101 (approved):
LGTM

One note is that those sed commands don't work out of the box on a mac. I don't think we need to delay anything for it but we can copy over the OS specific logic from other scripts at some point. i'll keep that task on my background list

nshaw:
@joewhite101 thanks. Good to know on the sed commands. At this point that will only impact OpenShift landing pages since the other flavors use the ent CLI but would be good to improve it.

PR #262, opened 2021-01-26
Fix instructions on how to view swagger UI

raxkaynan:


nshaw:
For the record, apparently there's some confusion about whether the Swagger UI should be enabled in a quickstart. This will be discussed in the Arch meeting on Friday and then we'll figure out what to do with the docs.

nshaw (commented):
@raxkaynan , this shouldn't be necessary. In a 6.3.0 quickstart, swagger isn't enabled by default. I'm going to add instructions on how to enable it, and then these steps are correct for me.

nshaw:
I think we can close this PR but please let me know if you disagree.

nshaw:
See https://github.com/entando/entando-docs/pull/266

PR #264, opened 2021-01-27
ENDOC-57 Improve redirects from landing pages so the 404 page doesn't display

nshaw:
This is visible on  my GH Pages - https://nshaw.github.io/docs/, https://nshaw.github.io/docs/getting-started, and https://nshaw.github.io/tutorials. If you watch closely you'll see a message about the redirect but I think that's preferable to a 404 (even temporary) and a full Vue reload.


joewhite101 (approved):
LGTM

PR #283, opened 2021-03-22
Update ecr-bundle-details.md

GermanoGiudici:
fix based on pda bundle content


nshaw (changes requested):
Could you please update docs/next/docs/ecr/ecr-bundle-details.md with the same change?  Until next diverges from v6.3 enough we're trying to update both. The next version will be used to create v6.4 at some point.

nshaw (approved):

PR #313, opened 2021-06-21
Update standard demo tutorial

nshaw:


nshaw:
Sorry about the whitespace/format changes. Tried to reverse that in the last edit but it was a straight copy from 6.3.2 so not sure how it happened.

nshaw:
@joewhite101 Might be a little early to rename from Examples to Solutions but figured I'd get the move started. We should be able to pull the existing PDA plugin tutorial down to that section as soon as it's working again and we add install instructions.

joewhite101 (approved):
LGTM

PR #324, opened 2021-09-27
missing steps for EKS install

ggros:
- Add note about IAM Role
- Move the create namespace
- Apply the operator config


nshaw:
@ggros Thanks for the fixes. I'll make a couple tweaks and update this on 6.3.2 as well and get this merged.

PR #325, opened 2021-09-27
Add the first to generating the access token

avdev4j:


nshaw:
@avdev4j One general comment. I know the screenshots make perfect sense for full clarity and they work well in blog posts which are a snapshot in time. However, in document, they age very quickly (as the tools themselves change) and the more images we use the harder it is to keep them in sync. We have a hard time keeping our own product snapshots in sync much less external tools so usually fewer images is safer. There's no right or wrong answer on how many but just keep in mind the maintenance cost.

PR #335, opened 2021-11-03
ENDOC-370 no support for M1 installation

Lyd1aCla1r3:
ENDOC-370 No Entando support for Apple's M1 chip noted in Getting Started and the Known Issues section of Release Notes

### DIFF
--- a/vuepress/docs/v6.3.2/docs/getting-started/README.md
+++ b/vuepress/docs/v6.3.2/docs/getting-started/README.md
@@ -4,7 +4,12 @@ sidebarDepth: 2
 
 # Getting Started with Entando
 
+::: warning
+**Mac:** Support for the Apple M1 platform is a work in progress and not offered by Entando 6.
+:::
+
 You have two options for getting started with Entando.
+
 1. [Automatically install Entando via the Entando command-line interface (CLI)](#automatic-install). This is the fastest way to start up an Entando application in Kubernetes.
 2. [Manually install Entando step by step](#manual-install). This is useful if you're preparing a shared cluster rather than a local developer environment, the CLI defaults don't meet your specific needs, or if you want to customize the deploy itself.
 
--- a/vuepress/docs/v6.3.2/docs/getting-started/README.md
+++ b/vuepress/docs/v6.3.2/docs/getting-started/README.md
@@ -5,11 +5,10 @@ sidebarDepth: 2
 # Getting Started with Entando
 
 ::: warning
-**Mac:** Support for the Apple M1 platform is a work in progress and not offered by Entando 6.
+**Mac:** Entando 6 is not currently compatible with the Apple M1 ARM64 architecture found in some newer Macs.
 :::
 
 You have two options for getting started with Entando.
-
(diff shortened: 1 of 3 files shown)


Lyd1aCla1r3:
Notes were confined to 6.3.2 pending approval; verbiage is succint and generalized to minimize and relieve focus from incompatibilities, and due to my inability to discuss the issue in great detail. Changes will be extended to Next once formatting/message has been ok'd or reworked.

jyunmitch (commented):
I like the second statement in the releases Readme. I would change the second sentence to read "Solutions are under investigation."

vuepress/docs/v6.3.2/docs/getting-started/README.md
```diff
@@ -4,7 +4,12 @@ sidebarDepth: 2
 
 # Getting Started with Entando
 
+::: warning
+**Mac:** Support for the Apple M1 platform is a work in progress and not offered by Entando 6.
```
nshaw, line 8: I had to stare at it for a bit, "not offered" doesn't feel quite right. We do offer enterprise support for the Entando platform in production which is a different thing altogether.  Maybe "Entando 6 is not currently compatible with the Apple M1 ARM64 architecture found in some newer Macs."

nshaw: I think this is closer to your release note which I do like. This doesn't promise too much. ;)

vuepress/docs/v6.3.2/docs/releases/README.md
```diff
@@ -100,5 +100,8 @@
   - There is  a known issue with the current nginx ingress that can be worked around by using the `singleHost` option for the configuration of your Entando Applications and binding TLS to that single host path
     - The issue will be solved in the next minor or patch release (6.4.0 or 6.3.3)
 
+### Apple M1
```
nshaw, line 103: This reads well, nice job. Minor suggestion, do we want the known issues to be alpha-sorted - Apple M1, then Entando Component Generator, then GKE?

nshaw (changes requested):

vuepress/docs/v6.3.2/docs/getting-started/README.md
Lyd1aCla1r3: Good call! You highlight a distinction that hadn't occurred to me - there are many flavors and stages of implementation across different sectors. Support in production is far different from a consumer use case where a customer is struggling with local installation and doesn't have an FAE resource.

vuepress/docs/v6.3.2/docs/releases/README.md
Lyd1aCla1r3: @nshaw Any opinion on omitting "issue" as @jyunmitch suggested?

nshaw: I like @jyunmitch's suggestion, bit more concise, but both read fine to me.

nshaw (approved):

PR #367, opened 2021-12-15
ENDOC-397 Replaced 61 occurrences across 35 files

Lyd1aCla1r3:
ENDOC-397 Update links to entando-de-app


nshaw (approved):
Probably could have just done next but probably no harm in doing the others since you're already made the changes.

Lyd1aCla1r3:
in vscode i did a replace all :)

PR #368, opened 2021-12-15
ENDOC-391 Added note to Known Issues

Lyd1aCla1r3:
Add a release note re docker hub numeric account names breaking deployment of plugins


nshaw (changes requested):
Looks good, let's add it to next and we're set.

Lyd1aCla1r3:
next doesn't have an analogous release notes section... should we create one?

nshaw:
> next doesn't have an analogous release notes section... should we create one?

Right, good call. I guess let's just leave it there. It probably deserves a permanent reference at some point but I'm not sure how it should be handled.

nshaw (approved):

PR #371, opened 2021-12-16
Update hello-world.md

jtydhr88:
There is no CONFIGURE button, according to context, it should be Design button


jyunmitch (commented):
Thanks for the help! This one was merged in a different PR because additional files had to be updated. You should have gotten an update on your other PR(370) as that got merged.  

Thanks again and hope you have a good day!

PR #373, opened 2021-12-17
Update identity-management.md

nshaw:


nshaw:
I think we should default to the quickstart settings rather than making the user edit this every time. And I'm tired of modifying this every time I setup a quickstart... ;)

jyunmitch (approved):

PR #383, opened 2022-01-07
ENDOC-405 Add instructions to configure a different storage class to Azure doc

Lyd1aCla1r3:
Note on configuring storage class. Are the format and location of this comment appropriate?


nshaw (changes requested):
Apologies, moving the target - let's add this to 6.3.2 as well. Looks good with that change.

jyunmitch (approved):

nshaw (approved):

PR #396, opened 2022-01-24
Endoc 393 Navigation Reorganization

jyunmitch:


Lyd1aCla1r3:
much better! :) i really think we need to group the busy left nav subsections into general, simply titled categories. reading through the current lists of long titles is a bit of burden and not very illuminating. for example, in Tutorials -> Compose an Application, can we have a "Content" subsection for the content pages?

vuepress/docs/.vuepress/next.js
```diff
@@ -1,69 +1,129 @@
...
+                    },
+                    {
+                        title: 'Bundle and Component Descriptors',
+                        path: path  + 'curate-share/ecr-bundle-details.md'
                     },
                     {
-                        title: 'PDA Architecture',
-                        path: path + 'concepts/pda-architecture.md'
+                        title: 'Filtering Bundles',
+                        path: path  + 'curate-share/ecr-bundle-filters.md'
```
nshaw, line 55: I think we can keep the paths shorter. They should be human readable but shorter is better for finding them later - and less likely to go completely out of date if we change the full title later. Here I would shorten to curate-share to curate.  Remember, these are visible in the URL so savvy users do notice them.

vuepress/docs/.vuepress/next.js
```diff
@@ -1,69 +1,129 @@
...
+                title: 'Compose an Application',
                 children: [
-                    path  + 'component-generator/component-gen-overview.md',
-                    path  + 'component-generator/blueprint-features.md',
-                    path  + 'component-generator/component-gen-tech.md',
-                    path  + 'component-generator/component-gen-customize.md',
-
+                    {
+                        title: 'Welcome Wizard',
+                        path: path + 'compose-an-app/welcome-wizard.md'
```
nshaw, line 84: Same, shorten "compase-an-app" to "compose"

vuepress/docs/.vuepress/next.js
```diff
@@ -121,188 +173,232 @@ module.exports = {
...
-                    path  + 'cms/content-attributes.md',
-                    path  + 'cms/content-templates-tutorial.md',
-                    path  + 'cms/digital-assets-tutorial.md',
-                    path  + 'cms/creating-protected-resources.md'
+                    {
+                        title: 'Micro Frontends',
+                        children: [
+                            {
+                                title: 'Introduction',
+                                path: path + 'tu-create-components/mfe/'
```
nshaw, line 264: I would definitely drop the "tu"

vuepress/docs/.vuepress/next.js
```diff
@@ -121,188 +173,232 @@ module.exports = {
...
                 ]
             },
             {
-                title: 'Configuration and Operations',
+                title: 'Configure and Customize an Application',
                 children: [
                     {
                         title: 'Default Database',
-                        path: path + 'devops/default-database.md',
+                        path: path + 'configure-customize/default-database.md',
```
nshaw, line 368: And shorten "configure-customize", maybe just back to "devops". Even though we don't use the term in the nav, it's more descriptive in a word for what is included here.

vuepress/docs/.vuepress/next.js
```diff
@@ -1,69 +1,129 @@
...
                     {
-                        title: 'Overview',
-                        path: path  + 'getting-started/',
+                        title: 'Introduction',
+                        path: path,
                     },
                     {
-                        title: 'Welcome Wizard',
-                        path: path + 'getting-started/welcome-wizard.md'
+                        title: 'Learn About Entando',
```
nshaw, line 12: I don't think Learn About Entando is right here - since this is the actual getting started install guide.

vuepress/docs/.vuepress/next.js
```diff
@@ -121,188 +173,232 @@ module.exports = {
...
-                        path: path  + 'micro-frontends/authentication.md',
-                    },
+                        title: 'Tutorials Learning Paths',
+                        path: path, 
+                    },
+                    {
+                        title: 'Install Entando',
+                        children: [
+                            {
+                                title: 'On Red Hat OpenShift',
```
nshaw, line 208: As we discussed let's drop the "On" in each one, so we'll get "Red Hat OpenShift", "Red Hat OpenShift using the OperatorHub", etc.

vuepress/docs/next/docs/consume-an-app/accessibility.md
```diff
@@ -41,7 +41,7 @@ design systems used by Entando clients include:
 * [Bootstrap Italia](https://github.com/italia/bootstrap-italia) - a Bootstrap 4-based frontend theme that implements
   the Italian Design Guidelines for public websites.
 
-Asessing web accessibility is important throughout the life of a project. There are many tools available in this area. A
+Assessing web accessibility is important throughout the life of a project. There are many tools available in this area. A
```
nshaw, line 44: Ouch, I wrote that one...

vuepress/docs/.vuepress/next.js
```diff
@@ -1,69 +1,129 @@
 module.exports = {
     docsSidebar: function(path) {
         return [
-            {
-                title: 'Introduction',
-                path: path,
-            },
             {
                 title: 'Getting Started',
```
nshaw, line 5: I feel like making the section content display by default is more usable, as we discussed. It also simplifies the left nav a bit. I tried this for the docs in my local:
```
               title: 'Overview',
                path: path,
                children: [
                    {
                        title: 'Get Started',
                        path: path + 'getting-started/',
                    },
                    {
                        title: 'Concepts',
                        path: path + 'getting-started/concepts-overview.md',
                    },
                ]
 ```

I think we want the same thing for the tutorials.

What I'm uncertain about is where we explain the Create-Curate-Compose-Consume idea in a way that it will get seen. We might need it at the top-level of docs and tutorials page, for docs definitely ahead of the diagram explaining the Entando architecture. We want to explain what Entando accomplishes. rather than how it does it, at least for the first impression.

**Review by nshaw — CHANGES_REQUESTED** [review]

File: vuepress/docs/next/docs/create/create.md [hunk 9]
```diff
@@ -0,0 +1,20 @@
+---
```
jyunmitch, line 1: All the text for this new file and for How to use this website is temporary. It is just to give you a feel for how it will feel and what might appear on the page. Don't waste your time reviewing it!

vuepress/docs/next/docs/getting-started/website.md
```diff
@@ -0,0 +1,26 @@
+---
```
jyunmitch, line 1: All the text for this new file and for Create Components landing page is temporary. It is just to give you a feel for how it will feel and what might appear on the page. Don't waste your time reviewing it!

vuepress/docs/.vuepress/next.js
```diff
@@ -3,67 +3,115 @@ module.exports = {
...
             {
                 title: 'Introduction',
-                path: path,
+                children: [
+                    {
+                        title: 'Getting Started',
+                        path: path + 'getting-started/',
+                    },
+                    {
+                        title: 'Entando Architecture',
```
nshaw, line 12: I played with this a bit and I think we're really close. The only thing that's bugging me is that the Entando Architecture page is the default and it's second in the list. Thinking it through, that is the current default page if you just come straight to /docs. I think that still makes sense but the name isn't quite right then. I'm thinking we should make it first in the list, rename it to Overview.  Then we have Overview, Getting Started, Concepts in the list.

nshaw (changes requested):
Just one comment for us to discuss. We need to figure out how to deal with the merge conflicts too...

nshaw:
Replaced by the new PR

PR #400, opened 2022-01-28
Endoc 393 nps

nshaw:


vuepress/docs/.vuepress/next.js
```diff
@@ -3,67 +3,115 @@ module.exports = {
...
-                        path: path + 'concepts/accessibility.md'
+                        title: 'Customizing The Component Generator',
+                        path: path + 'create/component-gen-customize.md',
+                    },
+                ]
+            },
+            {
+                title: 'Curate and Share Components',
+                children: [    
+                    {
```
jyunmitch, line 45: this link has to be removed, otherwise the expand/contract won't work. I think it has to do with the fact that entando-hub.md (not hub.md) is in the tutorial directory

Lyd1aCla1r3: sorry! i had changed hub.md to entando-hub.md - both tutorial file names + the references to them in the js files

jyunmitch: no problem, it's not the name of the file that's at issue anyways. I don't think we can cross from docs to tutorials in the sidebar.

nshaw: Let's discuss on the call. That was one I was unclear about.

jyunmitch (approved):

PR #404, opened 2022-01-31
ENDOC-416 Create New Entando Component Manager page

Lyd1aCla1r3:
missing images


vuepress/docs/next/docs/compose/ecm-overview.md
```diff
@@ -0,0 +1,20 @@
+# Entando Component Manager
+
+​​An Entando Application is composed of the Entando App Builder, Entando App Engine, and Entando Component Manager. The Entando Component Manager (ECM) is a service that links the [Entando Component Repository](ecr-overview.md) (ECR) of the App Builder to the core application instance. It appears as `quickstart-cm-deployment` in the Kubernetes pod list:
+
+![pods.png](./img/pods.png)
+
+The ECM communicates with the Entando Kubernetes service to populate the ECR with the bundles deployed as [Custom Resources](../consume/custom-resources.md) in the Entando cluster. These bundles are available to the application and can be managed from within the ECR.
```
nshaw, line 7: "deployed as... in the Entando cluster" change to "available as... in the Entando namespace. These bundles can be installed in the application and managed from within the ECR."

Not sure why I think that reads better but part of it is you don't deploy a Custom Resource - even though that's exactly the term we use in the UI for a bundle coming from the hub. It just feels wrong when it's said directly about a CR.

vuepress/docs/next/docs/compose/ecm-overview.md
```diff
@@ -0,0 +1,20 @@
+# Entando Component Manager
+
+​​An Entando Application is composed of the Entando App Builder, Entando App Engine, and Entando Component Manager. The Entando Component Manager (ECM) is a service that links the [Entando Component Repository](ecr-overview.md) (ECR) of the App Builder to the core application instance. It appears as `quickstart-cm-deployment` in the Kubernetes pod list:
+
+![pods.png](./img/pods.png)
+
+The ECM communicates with the Entando Kubernetes service to populate the ECR with the bundles deployed as [Custom Resources](../consume/custom-resources.md) in the Entando cluster. These bundles are available to the application and can be managed from within the ECR. 
+
+<img src="./img/ecm-flow.png" width="800" height="597.73">
```
nshaw, line 9: If we're going to include the numbers, maybe we should explain the steps. 1) A user goes to the ECR in the browser, 2) a request is made to the digital-exchange endpoint, 3) which is served by the Component Manager, 4) which calls the k8s-service to return the list of available bundles in the namespace.  A similar flow occurs when bundles are installed or uninstalled, with additional work done by the Operator which watches for events on the appropriate resources.

nshaw (changes requested):

vuepress/docs/next/docs/compose/ecm-overview.md
jyunmitch, line 7: minor point, in the first line, 'Entando Kubernetes service' >> can we drop the Entando? The first paragraph has so many Entando's, which was necessary, wondering if it's absolutely needed here?

Lyd1aCla1r3: i'm under the impression that this kubernetes service has been modified or integrated to be entando-specific. if that's the case, it *might* be important to note. i left it in to err on the side of caution, but i agree that we should drop the "Entando" if possible. @nshaw?

Lyd1aCla1r3: removed "Entando"

vuepress/docs/next/docs/compose/ecm-overview.md
```diff
@@ -1,15 +1,26 @@
...
 
-![pods.png](./img/pods.png)
+![pods.png](./img/pods.png) 
 
-The ECM communicates with the Entando Kubernetes service to populate the ECR with the bundles deployed as [Custom Resources](../consume/custom-resources.md) in the Entando cluster. These bundles are available to the application and can be managed from within the ECR. 
+The ECM communicates with the Kubernetes service to populate the ECR with the bundles available as [Custom Resources](../consume/custom-resources.md) in the Entando namespace. These bundles can be installed in the application and managed from within the ECR.
 
 <img src="./img/ecm-flow.png" width="800" height="597.73">
 
-The Component Manager provides functionality to build and organize micro frontends and widgets from within the App Builder. It also manages the connections between an application and the installed plugins.
```
jyunmitch: This summary works as a good intro to the steps below.

vuepress/docs/next/docs/compose/ecm-overview.md
```diff
@@ -1,15 +1,26 @@
...
 <img src="./img/ecm-flow.png" width="800" height="597.73">
 
-The Component Manager provides functionality to build and organize micro frontends and widgets from within the App Builder. It also manages the connections between an application and the installed plugins. 
+In the flow pictured above:
+1. A user visits the ECR in the browser
+2. A request is made to the digital-exchange endpoint
+3. The Component Manager serves the request
+4. The k8s-service is queried to return the list of available bundles in the namespace
 
+A similar process occurs when bundles are installed or uninstalled, except the [Entando Operator](../consume/operator-intro.md) performs additional event listening on the appropriate resources.
```
jyunmitch, line 23: I think I've seen it as a 'listening event'

nshaw (changes requested):
In the interest of time I made a few edits/suggestions here: https://github.com/entando/entando-docs/pull/408. Take what you like there and then let's get this into 6.3.2 and call it good.

nshaw:
Closing in favor of a new PR

PR #405, opened 2022-02-01
ENDOC-415 Create New App Builder page

Lyd1aCla1r3:


vuepress/docs/next/docs/compose/app-builder.md
```diff
@@ -0,0 +1,37 @@
...
+* Install component bundles from the Entando Component Repository
+
+* Append or update applications
+
+* Use micro frontends to build pages from modular, editable content
+
+* Deploy standalone packaged business capabilities
+
+* Deliver standardized UX design with page and content templates 
+
```
jyunmitch, line 27: Should we mention PDA functionality or being able to test and view your application right from here

vuepress/docs/next/docs/compose/app-builder.md
```diff
@@ -0,0 +1,37 @@
...
+
```
jyunmitch, line 28: how about a picture of the page editing section, where you have the list of components to add on the side. I always liked that capability

vuepress/docs/next/docs/compose/app-builder.md
Lyd1aCla1r3: i'd say no to PDA functionality... the default ECR is empty; the PDA plugin has to be installed. good point about preview!

vuepress/docs/next/docs/compose/app-builder.md
```diff
@@ -0,0 +1,43 @@
+---
+sidebarDepth: 2
+---
+
+# Entando App Builder
+
+The Entando App Builder plays a key role in the composition of Entando Applications. It is a feature-rich, low-code user interface to customize and manage components. It also offers a Dashboard and a modular CMS to streamline the design and build process. 
+
+<img src="./img/app-builder.png" width="800" height="393.76">
```
nshaw, line 9: Two comment on screenshots. 1) usually I try to screenshot just enough to show the functionality. In this case we're showing the App Builder so what you have is fine but typically we can only show a smaller area. Just mentioning that for next time. 2) Maybe VSCode is automatically adding the width/height but vuepress is pretty good about displaying the images without the width/height settings. In fact, in my testing Vuepress does a better job of displaying the images when I resize the browser window if the width/height is NOT set. Please try that and if you agree let's remove the width/height for any new images unless there's a compelling reason to do otherwise.

nshaw (changes requested):
One suggestion and then let's get that added to 6.3.2

vuepress/docs/next/docs/compose/app-builder.md
Lyd1aCla1r3: 1) agree and noted. moving forward only the area of interest will be documented in the image, especially if the surrounding UI is subject to change. 
2) i'll remove the width/height assignments; i misunderstood and was making those explicit after our hub discussion. all images are fed into tinypng to reduce file size.

nshaw: Per #2, cool. I think we had a special case on the hub page where it looked weird to have the images too big. I think what we have here is closer to our default case, basically allowing Vuepress to do its thing with a minimum of code.

nshaw (changes requested):
Looks good now so just need to add to 6.3.2

nshaw:
Closing in favor of the new PR

PR #409, opened 2022-02-03
ENDOC-418B change dev.entando.org to developer.entando.com

jyunmitch:


nshaw (approved):
@jyunmitch looks good. The extra search results in my local were in the generated target folder for the old docs. Those files aren't controlled and you wouldn't see them unless you ran the asciidoc generator. So we're good here.

PR #418, opened 2022-02-09
Endoc 419

jyunmitch:
Please ignore the next.js and concepts page in this PR! Because I was working on both of them, I wasn't as careful as I should have been and didn't notice it was added. I couldn't revert because it was commited with too much of the changes here. Sorry. 

But it should not cause a conflict cuz the files should be the same and the most current.

Focus on the first 4 commits as they are pertinent ones for the PR


nshaw (approved):
I used the web editor to deal with the merge conflict. I think all that is left is some whitespace changes on development-concepts but please double check.

PR #420, opened 2022-02-10
ENDOC-426 Copy next landing-page into v6.3.2. Correct a footer style.

nshaw:
1. Straight copy of the landing-page into v6.3.2
2. Fixed the footer style so the Edit this page link shows up on all but the home/landing pages.


nshaw:
I did publish this to staging already to double-check the styles and because I wanted to show Manuela the newest content.

jyunmitch (approved):

PR #432, opened 2022-02-22
ENDOC-429

Lyd1aCla1r3:
standardize client role assignment by pointing to section on revamped identity management page and specifying only which client role and available role to choose

### DIFF
--- a/vuepress/docs/next/tutorials/solution/entando-hub.md
+++ b/vuepress/docs/next/tutorials/solution/entando-hub.md
@@ -57,8 +57,8 @@ Alternatively, restarting the quickstart-server pod (which contains the Entando 
 :::
 
 5. Setup permissions to configure the service:
-   - Login to Keycloak as an admin. To manage the required Keycloak instance, see [Entando Identity Management -- Keycloak](../../docs/consume/identity-management.md#logging-into-your-keycloak-instance).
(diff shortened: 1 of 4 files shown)


vuepress/docs/next/tutorials/solution/entando-hub.md
```diff
@@ -58,7 +58,7 @@ Alternatively, restarting the quickstart-server pod (which contains the Entando
 
 5. Setup permissions to configure the service:
    - Login to Keycloak as an admin. To manage the required Keycloak instance, see [Entando Identity Management -- Keycloak](../../docs/consume/identity-management.md#logging-into-your-keycloak-instance).
-   - Go to `Users → admin → Role Mappings`, select `entandopsdh-entando-hub-catalog-server` as the `Client Role`, then add `eh-admin` to `Assigned Roles`.
+   - Assign client roles by following [these instructions](../../docs/consume/identity-management.md#authorization). Select `entandopsdh-entando-hub-catalog-server` from `Client Roles` and `eh-admin` from `Available Roles`.
```
jyunmitch, line 61: we have used link references with 'these instructions' or 'here' before but wondering if it's better to be more clear with something like "the  Identity Management instructions". We should discuss for style guide, it's a relatively straightforward issue.

Lyd1aCla1r3: totally agree. i used a vague reference because i didn't want to introduce confusion or redundancy by calling out the entando identity management page 2x in a row, for different purposes, but i think i can rephrase to specify the sections; that should avoid any issues and replace language that doesn't add value

jyunmitch (approved):

PR #435, opened 2022-02-25
ENDOC-431b Update the groups descriptor documentation

Lyd1aCla1r3:


vuepress/docs/next/docs/curate/ecr-bundle-details.md
```diff
@@ -1,5 +1,6 @@
 ---
 sidebarDepth: 2
+redirectFrom: /v6.3.2/docs/ecr/ecr-bundle-details.html
```
jyunmitch, line 3: Are we doing redirects for some documents from 6.3.2 to next?

vuepress/docs/next/docs/curate/ecr-bundle-details.md
```diff
@@ -239,9 +240,11 @@ Here is an example of a Page Template descriptor:
       </html>
 
 ## Page
-This descriptor enables a page to be created and published via a bundle. Page status can be `published` or `draft`. The widget section can be used to fully configure a page layout.
+This descriptor enables a page to be created and published via a bundle. Page status can be `published` or `draft`. The Widget section can be used to fully configure a page layout. 
 
-**Page descriptor.yaml.**
+Groups in a page descriptor are configured by `ownerGroup` and `joinGroups`. The `ownerGroup` property assigns the group of users who can manage the entity in the AppBuilder. The `joinGroups` property defines who can view or access the entity. A group designation of "free" signifies a public entity.
```
jyunmitch, line 245: this part is confusing>> "The `ownerGroup` property assigns the group of users who can manage the entity in the AppBuilder. The `joinGroups` property defines who can view or access the entity." Who is assigning what to whom? Isn't it more about if you assign a user to 'ownerGroup', they can manage the entity in the AppBuilder... same with joinGroups defining who can view or access? I think property is misleading maybe.

nshaw: Maybe "assigns" -> "specifies". And then -> `joinGroups` specifies who can view or access the page. For example, setting the `ownerGroup` to `free` means anyone with access to the AppBuilder can manage the page, whereas setting the `joinGroup` to `free` means any end user can see view the page in the application.

vuepress/docs/next/docs/curate/ecr-bundle-details.md
```diff
@@ -410,12 +412,16 @@ For more details on Content Type properties, refer to the [Content Type document
 ## Content
 This descriptor enables content to be created and optionally published via a bundle, according to the `status` property. The content ID is optional and enables linking from other components, like Content Widget. It can be autogenerated or explicitly declared.
 
-**Content descriptor.yaml.**
+Groups in a content descriptor are configured by the `mainGroup` and `groups` properties. A group designation of "free" signifies a public entity.
```
nshaw, line 415: '...by the `mainGroup` and `groups` properties. A group designation of "free" signifies a public entity.' -> 'by the `mainGroup` and `groups`, for the owner and join groups, respectively.  The `mainGroup` is for the users who can manage the content within the AppBuilder and `groups` for those users who can view the content.`

nshaw (changes requested):

vuepress/docs/next/docs/curate/ecr-bundle-details.md
Lyd1aCla1r3: this is always subtle and tricky for me, because someone determines what/who a property or attribute consists of, but in turn that property/attribute controls the actions or access of what/who... agree 100% wrt "assigns" here tho

jyunmitch: I know in a broad sense, these can be viewed as properties or attributes but in these specific cases, aren't these groups and roles and they provide the attribute of who has what access

Lyd1aCla1r3: actually groups and roles are properties... properly identifying programming elements is not a strength of mine, but nathan defined them as such so i'm confident :P regardless i think your point of who is assigning what to whom is the important bit, like is the programming element assigning something to a user or a user being assigned to a programming element. sometimes it's obv and cut and dried, but other times there's some gray area. in this case we confirmed the properties aren't assigning anything, just specifying; an admin is assigning a value to the property that defines user access, etc.

vuepress/docs/next/docs/curate/ecr-bundle-details.md
Lyd1aCla1r3: mistake!

nshaw (approved):

PR #515, opened 2022-05-27
ENDOC-506 add M1 comment

jyunmitch:
Could not find a smaller font-size in the CSS file. Wouldn't mind if the text size was smaller. Should I try to add another class to the CSS file?


Lyd1aCla1r3 (approved):

nshaw (approved):
Let's just go with it.

PR #519, opened 2022-05-31
ENDOC-506 Add M1 alt links to get started page

jyunmitch:


vuepress/docs/v7.0/docs/getting-started/README.md
```diff
@@ -7,8 +7,8 @@ sidebarDepth: 2
 
 
 ::: warning
-**Mac:** Entando 7 is not currently compatible with the Apple M1 ARM64 architecture found in some newer Macs.
-:::
+If you have a MacBook with an M1 chip, please see our alternative [install guides](../../tutorials/#operations).
```
nshaw, line 10: Let's add a second sentence to provide a bit more information. "Entando 7 is not currently compatible with an ARM64-based architecture." That way we tell them why as well as what to do.

And do we need this on next too?

nshaw (changes requested):

PR #526, opened 2022-06-16
Update external-db.md

gp-ent:
Fix Oracle Dialect driver


nshaw (approved):
@jyunmitch or @Lyd1aCla1r3 please also add this to /next.

PR #539, opened 2022-07-21
ENDOC-528 Add OKD 4.8 work-around comment on PDA page

Lyd1aCla1r3:


vuepress/docs/next/tutorials/solution/pda-tutorial.md
```diff
@@ -56,6 +56,42 @@ ent ecr deploy --repo="https://github.com/entando-samples/entando-process-driven
 
 ## Configuration
 
+::: warning
+---
+
+<details><summary>Workaround for PDA connections on OpenShift 4.8</summary>
+
+The PDA Plugin connections cannot be configured for RedHat OpenShift 4.8 implementations of Entando 7 due to missing rules on the entando-plugin Role. To circumvent a connections error, add the following definition to `namespace-resources.yaml`:
```
nshaw, line 64: Minor change, my notes were not clear. "To circumvent a connections error, update the `entando-plugin` Role by adding the missing rules below."

I'm leaving it a little vague since you could do this with kubectl, k9s, or in the OpenShift UI like I did. This requires a bit of k8s knowledge of course.

I know for sure we need this on the 7.0 codebase but I'm unsure about 7.1 until we test. Maybe tweak this, add it to 7.0, and then add a note to our 7.1 TODOs to review later?

nshaw (changes requested):

vuepress/docs/v7.0/tutorials/solution/pda-tutorial.md
```diff
@@ -56,6 +56,42 @@ ent ecr deploy --repo="https://github.com/entando-samples/entando-process-driven
 
 ## Configuration
 
+::: warning
+---
+
+<details><summary>OpenShift 4.8 requires a workaround</summary>
+
+An incomplete role definition prevents Entando 7 from configuring the PDA Plugin connections on RedHat OpenShift 4.8. As a workaround, update the entando-plugin Role by adding the missing rules below.
```
jyunmitch, line 64: isn't this supposed to have the bit about circumventing the connection error? otherwise everything else looked fine.

Lyd1aCla1r3: i reworded it... "circumventing" (my word choice) sounded not quite right; it implies dodging something that's an inevitability. "workaround" is standard language for a quick and dirty - and temporary - fix. but explicitly calling out that something will error gives off an unfinished, amateur vibe, instead of finessing the inferences like in the revision.

jyunmitch: I agree that the mention of an error is not necessarily a good thing but sometimes I think the meaning starts to get a bit vague. In this case, it is understandable but not crystal clear.

jyunmitch (approved):

vuepress/docs/v7.0/tutorials/solution/pda-tutorial.md
Lyd1aCla1r3: good point

nshaw (approved):

PR #540, opened 2022-07-26
ENDOC-524 7.1 Ent CLI-create new API management page

Lyd1aCla1r3:


vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,72 @@
...
+    "serviceName": "ext-ms",
+    "bundle": "registry.hub.docker.com/my-org/my-entando-bundle"
+}
+```
+
+## API Claim Commands
+
+Common operations associated with API claims are detailed below.
+
+The `ent profile` command is available to manage and switch between the configurations of different Entando instances. Refer to `ent profile first-use-readme` for additional information.
```
nshaw, line 38: ent profile?

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,72 @@
...
+		serviceName: int-ms
+	- name: ext-api
+		type: external
+		bundle: registry.hub.docker.com/cecchisandrone/entando-simple-bundle
+		serviceName: ext-ms
+```
+
+- `name` and `type` correspond to the same fields in the API claim source descriptor
+- `pluginName` corresponds to `serviceName`
+- `bundleId` is a generated ID derived from the bundle
```
nshaw, line 72: Might be good to link this to the bundleId generation section.

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,72 @@
...
+    "name": "ext-api",
+    "type": "external",
+    "serviceName": "ext-ms",
+    "bundle": "registry.hub.docker.com/my-org/my-entando-bundle"
+}
+```
+
+## API Claim Commands
+
+Common operations associated with API claims are detailed below.
```
nshaw, line 36: Might be good to show an example of the full command (ent bundle api) somewhere?

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,72 @@
...
+    "serviceName": "int-ms"
+}
+```
+**External API claim**
+``` json
+{
+    "name": "ext-api",
+    "type": "external",
+    "serviceName": "ext-ms",
+    "bundle": "registry.hub.docker.com/my-org/my-entando-bundle"
```
nshaw, line 30: Open question: can this only reference other bundles within the same project/namespace? We might @w-caffiero-entando to comment but "external" sounds like it could be external to this bundle (as it appears) or external to this project/namespace - say to other projects or even outside kubernetes to other enterprise systems or even to external 3rd party services in the cloud. Based on the format of the bundles I suspect it's only the first case but we should  probably add a clarifying note on the scope.

nshaw (changes requested):

nshaw (approved):
Couple minor open questions but good for me.

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,72 @@
...
+   - `bundle`: The external bundle URL
+   - `service-name`: The name of a microservice in the external bundle. If `service-name` is not set, `api add-ext` will initiate an interactive mode where the user can select from available bundles and microservices.
+
+## API Claim YAML Descriptor Mapping
+
+When the YAML descriptor of a micro frontend is generated (i.e. via the `pack` command), the API claim mapping is similar to the following:
+
+``` yaml
+apiClaims:
+	- name: int-api
```
jyunmitch, line 59: what are these red bars?

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,72 @@
...
+
+## API Claim Commands
+
+Common operations associated with API claims are detailed below.
+
+The `ent profile` command is available to manage and switch between the configurations of different Entando instances. Refer to `ent profile first-use-readme` for additional information.
+
+| Command | Description
+| :- | :-
+| `api add [mfe-name] [claim-name]` | Add an internal API claim to an MFE
```
jyunmitch: [mfe-name] >> YOUR-MFE-NAME
[claim-name] >> YOUR-CLAIM-NAME

since these are place holders, I've been using the all caps format with YOUR attached, and no brackets.

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,70 @@
+---
+sidebarDepth: 1
+---
 
 
+# Overview
 
+An API claim informs the Entando Platform that a microfrontend (MFE) or microservice (MS) intends to use the API of another microservice. This request can be sent within a single bundle or across bundles. The abstraction of an API claim eliminates the need to define and manage API endpoints, both in local development and within a running instance.
```
jyunmitch, line 8: micro frontend

jyunmitch (approved):

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -1,4 +1,70 @@
...
+
+## API Claim Commands
+
+Common operations associated with API claims are detailed below.
+
+| Command | Description
+| :- | :-
+| `ent bundle api add [mfe-name] [claim-name]` | Add an internal API claim to an MFE
+| `ent bundle api add-ext [mfe-name] [claim-name]` | Add an external API claim to an MFE
+| `ent bundle api rm [mfe-name] [claim-name]` | Remove an API claim from an MFE
```
zonia3000, line 42: I would also add a note about the fact that all `ent bundle api` subcommands can be executed only if the user is attached to a running Entando (see `ent attach` commands), otherwise the CLI will produce an authentication error.

vuepress/docs/next/docs/getting-started/ent-api.md
zonia3000: In this context `external` means external to the bundle, but not to the Kubernetes namespace. Indeed, when you execute `ent bundle api add-ext` you have to be connected to an Entando instance and the command will fail if it sees that the service you are specifying in the API claim doesn't exist on the cluster. The external service has to be deployed before adding its API claim.

zonia3000: When you execute the command without the flags it searches the available services on the instance, so that you can select one of them interactively:
image
The command also displays that you are running it on a given instance.

vuepress/docs/next/docs/getting-started/ent-api.md
nshaw: @jyunmitch Minor, might be better to write `ent attach-*` since the exact commands are attach-vm, attach-kubeconfig, attach-kubectx

nshaw (approved):

vuepress/docs/next/docs/getting-started/ent-api.md
```diff
@@ -50,21 +50,40 @@ Common operations associated with API claims are detailed below.
...
-		serviceName: ext-ms
+## Microservice URL Retrieval
+
+To render an MFE installed from a bundle, Entando injects a JSON object containing configuration details inside the "config" attribute of a custom HTML element, e.g. `<my-mfe config="{ ... }" />`.
+
+**Installed Bundles**
+
+To retrieve the URL of an MS declared through an API claim, add `systemParams.api[claimName].url` to the MFE JavaScript code so it can access that element of the config object. The `claimName` was chosen by the user when the API claim was defined.
+
+Refer to the Entando demo bundles for alternative retrieval methods. These techniques are more robust and consider cases where the parent object is not defined.
```
jyunmitch, line 70: these techniques>> "These" techniques are referring to the alternative retrieval method ones, right? just checking

jyunmitch (approved):

vuepress/docs/next/docs/getting-started/ent-api.md
Lyd1aCla1r3: yes, but i'll combine those 2 sentences to clarify

PR #550, opened 2022-08-19
ENDOC-typo correct authentication in sidebar nav

jyunmitch:
ironic the branch name has a typo too.

the nav had authentification, instead of authentication as the page is called. Probably my mistake to begin with.


Lyd1aCla1r3 (approved):
"authentification"... i like it!

nshaw:
Wow, how did we miss that? Confirmation bias!

nshaw (approved):

PR #556, opened 2022-08-26
ENDOC-561 7.1 simple mfe react app

Lyd1aCla1r3:


vuepress/docs/next/tutorials/create/mfe/widget-configuration.md
```diff
@@ -98,7 +98,7 @@ npm start
 npm run build
 ```
 
-6. Load the updated `my-widget` files into Entando as was done for the [React MFE tutorial](./react.md#upload-the-react-files)
+6. Load the updated `my-widget` files into Entando as was done for the [React MFE tutorial](../../../../v7.0/tutorials/create/mfe/react.md#upload-the-react-files)
```
nshaw, line 101: This should be a relative link so it points to the current version, rather than directly to v7.0

nshaw (changes requested):

vuepress/docs/next/tutorials/create/mfe/widget-configuration.md
Lyd1aCla1r3: @nshaw this is a tricky one because this section no longer exists in the current version, and it's not just a matter of renaming - we killed the content. i didn't look at the affected page closely enough to get a sense of how we might also modify it for v5, if at all, so just pointed back to 7.0 instead of removing the link reference and listing all the steps explicitly.

nshaw: @Lyd1aCla1r3 ah, understood. I have some ideas for it based on the Tableau widget conversion I'm working on so we can discuss on our next call.

vuepress/docs/next/tutorials/create/mfe/react.md
```diff
@@ -1,169 +1,123 @@
...
+3. Deploy the bundle to your Entando Application:
+   ``` sh
+   ent bundle deploy
+   ```
+   Your bundle will appear in the Local Hub of your Entando instance, accessible from `App Builder` → `Hub`, and show a status of DEPLOYED.
 
-- `my-widget/static/css`
-- `my-widget/static/js`
-- `my-widget/static/media`
+4. Install the bundle in your Entando Application from `App Builder` → `Hub` or with the following command:
```
jyunmitch, line 108: this is awkward and confusing. It also reads like "install the bundle in your entando application from app builder to the Hub" I would break it up or edit

nshaw (approved):

PR #564, opened 2022-09-06
ENDOC-561 Cover for Last Week Lydia

Lyd1aCla1r3:


vuepress/docs/next/tutorials/create/mfe/react.md
```diff
@@ -90,6 +92,7 @@ You’re now using a custom element to display a React app.
 
 1. From the root bundle directory, generate the Docker image:
    ``` sh
+   cd simple-bundle
```
nshaw, line 95: I don't think you need this command. If you're in the MFE subdirectory, you would either use cd ../.. - or simply leave the comment that they should work from the root bundle directory.

nshaw (approved):

nshaw (approved):

vuepress/docs/next/tutorials/create/mfe/react.md
```diff
@@ -13,23 +13,25 @@
    ent bundle init simple-bundle
    ```
 
-2. Add an MFE to the bundle project structure:
+2. From the root bundle folder, add an MFE to the bundle project structure:
```
jyunmitch, line 16: adding an MFE is not really adding it to the structure, so this seems misleading, like it's only a structural thing when it adds a folder but also adds it to the entando.json

vuepress/docs/next/tutorials/create/mfe/react.md
```diff
@@ -13,23 +13,25 @@
...
 [Create React App](https://create-react-app.dev/) allows you to generate a simple app in seconds. 
 
-1. From the bundle root folder, create a React app in `/simple-bundle/microfrontends/`: 
+1. Create a React app in `/simple-bundle/microfrontends/`: 
    ``` bash
+   cd microfrontends
    npx create-react-app simple-mfe --use-npm
    ```
-   Use the bundle name you chose when adding an MFE to overwrite the empty bundle folder created in the previous step. 
+   Assign the same bundle name you chose when adding an MFE to overwrite that empty bundle folder.
```
jyunmitch, line 30: empty bundle folder>> empty MFE folder

vuepress/docs/next/tutorials/create/mfe/react.md
Lyd1aCla1r3: disagree, because structure pertains to both file contents and directory, like how code is written as well as file hierarchy. i'll remove "structure" to keep both implied without an extra word

PR #575, opened 2022-09-20
ENDOC-507 Widget communication

Lyd1aCla1r3:


vuepress/docs/next/tutorials/create/mfe/communication.md
```diff
@@ -215,30 +178,27 @@ customElements.define('subscriber-widget', SubscriberWidgetElement);
...
-    <subscriber-widget>
+    <subscriber-widget />
     ...
   </body>
 ```
 
 ### Display Custom Event
 
-Update `subscriber-widget/src/App.js`.
+1. Update `subscriber-widget/src/App.js` to display the event:
```
jyunmitch, line 201: again the same, are you supposed add the whole thing, bits of it, or what?

vuepress/docs/next/tutorials/create/mfe/communication.md
```diff
@@ -265,392 +223,51 @@ const widgetEvent = new CustomEvent('greeting', {
...
-### Create Environment File
-
-#### Publisher Widget
-
-1. Create an `.env` file in the project root for the `publisher-widget`.
-
-2. Open the `.env` file, and enter the `PUBLIC_URL` where the micro frontend will be hosted.
-
-Example:
+To add the publisher and subscriber MFEs to Entando, run the following commands from the root folder of each:
```
jyunmitch, line 234: each reads like it's referencing publish and subscriber, not the MFEs, even though indirectly they are mfes but I was thinking what publisher directory?

vuepress/docs/next/tutorials/create/mfe/communication.md
```diff
@@ -265,392 +223,51 @@ const widgetEvent = new CustomEvent('greeting', {
...
+3. Click `Edit`
 
-- `Page Template: 1 Column`
+4. In the `Settings` section, select a Page Template with more than one frame
 
-5. Click `Save and Configure`.
+5. Click `Save and Design`
 
-6. In the `WIDGETS` sidebar on the right:
+6. From the `Widgets` tab in the right sidebar, drag your publisher and subscriber widgets into `Frame 1` and `Frame 2`
```
jyunmitch, line 259: drag >> drag and drop
it doesn't matter which goes in which frame, right?

vuepress/docs/next/tutorials/create/mfe/communication.md
```diff
@@ -6,38 +6,21 @@ sidebarDepth: 2
...
-``` bash
-cd publisher-widget
-```
-
-``` bash
-npm start
-```
+## Modify the Publisher MFE
 
 ### Create Custom Event
```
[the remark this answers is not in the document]
Lyd1aCla1r3: the original didn't use articles in the sub headers

vuepress/docs/next/tutorials/create/mfe/communication.md
```diff
@@ -76,37 +59,33 @@ customElements.define('publisher-widget', PublisherWidgetElement);
 export default PublisherWidgetElement;
 ```
 
-- In the `CustomEvent` constructor, `detail` is the specific name to use in the event payload, as per the [DOM specification](https://dom.spec.whatwg.org/#interface-customevent).
-
-#### Import Custom Element
+   In the `CustomEvent` constructor, `detail` denotes the specific name to use in the event payload per the [DOM specification](https://dom.spec.whatwg.org/#interface-customevent).
 
-Update `publisher-widget/src/index.js`.
+2. To import the custom element, update `publisher-widget/src/index.js`:
```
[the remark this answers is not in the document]
Lyd1aCla1r3: no. update means replace the contents. in other tutorials we've edited this language to make that explicit but i feel like the intent is understood so i left as-is. to expand upon it seems unnecessary because if you're familiar with the file content update=replace should be obvious

vuepress/docs/next/tutorials/create/mfe/communication.md
Lyd1aCla1r3: dragging the widget into the frame performs a drag and drop

it doesn't. the way we read from left to right, top to bottom would map to publisher on left/top but that's not worth mentioning

Lyd1aCla1r3:
@nshaw what are your thoughts on adding some language to clarify that update means replace the contents of? there's one instance where update means change a snippet, which would require different clarifying language. the instructions were clear to me as-is, so i'd assume any dev following this tutorial wouldn't need it spelled out??

vuepress/docs/next/tutorials/create/mfe/communication.md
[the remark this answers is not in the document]
Lyd1aCla1r3: @jyunmitch struggling with this one... this tutorial is identified as intermediate wrt our learning paths... i asked nathan to weigh in

nshaw:
@Lyd1aCla1r3 I've definitely had people get confused on that one. It might be nice to use different language when you replace the whole file, versus adding/updating a section. It seems like this tutorial isn't quite as explicit as some tutorials in that sense.

nshaw (approved):

vuepress/docs/next/tutorials/create/mfe/communication.md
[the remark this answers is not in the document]
jyunmitch: noticed the headers below used them

vuepress/docs/next/tutorials/create/mfe/communication.md
jyunmitch: not everyone reads left to right. I would totally read this one to mean put it in any frame

vuepress/docs/next/tutorials/create/mfe/communication.md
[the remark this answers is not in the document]
Lyd1aCla1r3: i liked it better with the articles and started adding them in but then second-guessed it and reverted; i guess i missed a couple. changed my mind again tho, so back to using articles

vuepress/docs/next/tutorials/create/mfe/communication.md
Lyd1aCla1r3: true, all the more reason to omit specific placement from the instruction. placement has no impact on functionality and is not subject to any constraints; the user is free to place the widgets in the layout of their choice. because the widgets exhibit some causality, the user will likely place them according to convention for cause and effect

PR #578, opened 2022-09-23
ENDOC-511 Update tutorials to reflect changes to bundle/resource codes

Lyd1aCla1r3:


vuepress/docs/next/docs/curate/bundle-comparison.md
```diff
@@ -4,7 +4,7 @@ sidebarDepth: 2
 
 # Bundle Evolution on Entando 7.1
 
-Entando 7.1 implements a more modular approach to building apps with the **ent bundle** CLI orchestrating the management of project bundles. The Entando Bundle was restructured, its processes decoupled by component type, and the publishing transport method, streamlined. With these changes, Entando Bundles went from git-based to docker-based OCI bundles. Both methods are supported on Entando 7.1, and this page clarifies the differences between the two, along with the improvements in the new release.
+Entando 7.1 implements a more modular approach to building apps with the **ent bundle CLI** orchestrating the management of project bundles. The Entando Bundle was restructured, its processes decoupled by component type, and the publishing transport method was streamlined. With these changes, Entando Bundles went from git-based to docker-based OCI bundles. Both types are supported on Entando 7.1, and this page clarifies the differences between the two, along with the improvements in the new release.
```
jyunmitch, line 7: just a note: both methods and types are work in this sentence. But it's not just types that are supported in the specific sense, because the way the bundles are processed in the building, packing, etc. is different. The way it is processed is different too so methods is technically applicable too.

vuepress/docs/next/docs/curate/bundle-comparison.md
```diff
@@ -15,21 +15,22 @@ Some of the major improvements with docker-based bundles are:
...
-|API Claims| Manual configuration in MFE bundle | [Manage APIs with `ent bundle` commands](../getting-started/ent-api.md). MFEs can make API claims to MSs within the same bundle or namespace
-|Local Services |Blueprint includes a few services under src/main/docker. | Default services now include Keycloak, PostgreSQL, and MySQL. Additional services can be added with the [`ent bundle svc` commands](../getting-started/ent-svc.md).
-|MFE Custom UI| Add with manual scripts| Custom UI is auto-generated in FTL format
-|MFE Configuration| Manually defined in the FTL | Utilize context parameters, system parameters and the `ext` variable to pass information to widgets and bundles, or expose it in the APIs
-|| | Auto-generated FTLs with parameters defined in `entando.json` |
-|Bundle Thumbnail| Detailed inside the YAML bundle descriptor| Set with a JPG or PNG file named `thumbnail` in the bundle root folder, e.g. thumbnail.png
+|Bundle creation | [`ent prj` commands](../getting-started/ent-bundle.md#git-based-bundle-commands)| New modular [`ent bundle` commands](../getting-started/ent-bundle.md)
```
jyunmitch, line 27: Everything in the first column was capitalized because they are row headers to give it some emphasis.

Lyd1aCla1r3: the column is labeled "Attribute" and capitalizing the row entries implies that each of those terms is an attribute, which is incorrect

vuepress/docs/next/docs/curate/bundle-comparison.md
```diff
@@ -15,21 +15,21 @@ Some of the major improvements with docker-based bundles are:
...
-| Versioning |Bundle version set by Git tags |Bundle version set by Docker tags. For MFEs, the version defaults to 1.0.0 unless set in the [entando.json](bundle-details.md). 
-|Plugin Ingress| Generated by Entando or set in the plugin descriptor.yaml | Generated by Entando or customized in the [bundle descriptor](bundle-details.md#entando-bundle-conventions), entando.json 
-|Identifiers| Designated Bundle ID and Plugin ID | Bundles, MFEs, & Plugins have unique global IDs|
+| Bundle specifications |Defined with descriptor files | Built with [ent bundle CLI](../getting-started/ent-bundle.md) tool
+| Versioning |Bundle version set by Git tags |Bundle version set by Docker tags. For MFEs, the version defaults to 1.0.0 unless set in the [entando.json](bundle-details.md)
+|Plugin ingress| Generated by Entando or set in the plugin descriptor.yaml | Generated by Entando or customized in the [bundle descriptor](bundle-details.md#entando-bundle-conventions), entando.json
+|Identifiers| Designated Bundle ID and Plugin ID | Bundles, MFEs, & plugins have unique global IDs|
+|Bundle ID| Not included in component names | Suffix of bundle, plugin, widget, page and fragment names
```
jyunmitch, line 25: Not included in component names >> I don't think this is correct, it did have bundle Id, just not global & generated in a different way that would not necessarily be unique globally,

vuepress/docs/next/docs/curate/bundle-comparison.md
Lyd1aCla1r3: the subject is bundles, of which there are types. a bundle is not a method. we would need to rephrase / change the subject to make "methods" work

jyunmitch: the paragraph talks about bundles and how they are processed

Lyd1aCla1r3: methods doesn't work here. if anything, it would refer back to the transport method mentioned 2 sentences prior. more directly, the previous sentence concerns bundles. to use a synonym for the subject that refers to something other than a bundle in the next breath doesn't work.

jyunmitch: You have a point, with the previous sentence, it flows nicer. But the topic sentence of the paragraph is not bundles but how bundles are managed by the ent bundle cli. So both words work, 'type' flows better, while 'method' emphasizes the process.

jyunmitch (approved):

PR #579, opened 2022-09-26
ENDOC-566 7.1 Edits for loose threads

jyunmitch:


vuepress/docs/next/docs/getting-started/ent-bundle.md
```diff
@@ -10,7 +10,7 @@ In addition, this document describes the series of `ent ecr` commands that manag
 
 ## Entando 7.1 Bundle Development
 
-Beginning with Entando 7.1, the `ent bundle` command and its convenience methods introduce a streamlined process to govern the structure, files and management of Entando bundles. The bundle development lifecycle consists of 6 stages, each corresponding to a subcommand:
+Beginning with Entando 7.1, the `ent bundle` module introduces a streamlined process to govern the files, structure, and management of Entando bundles. The bundle development lifecycle consists of 6 stages, each corresponding to a subcommand:
```
Lyd1aCla1r3, line 13: `ent bundle cli` tool (per discussion)

structure is the broadest term in that list and encompasses files, so the list should lead with it

jyunmitch: My understanding is that ascending or descending is fine. In this case, because I wanted end with the management, I picked ascending. The thinking was that management is bigger than structure since it controls it.

Lyd1aCla1r3 (approved):

PR #591, opened 2022-10-14
ENDOC-589 setup v7.1

nshaw:


nshaw:
This is published to staging now. I'm going to take another look later and likely publish to prod this weekend.

PR #605, opened 2022-10-25
Endoc 526B

jyunmitch:
Had to create new branch because the original 526 did not have v7.1 at all


vuepress/docs/next/tutorials/curate/bundle-private-images.md
```diff
@@ -0,0 +1,85 @@
+---
+sidebarDepth: 2
+---
+
+# Install Bundle from a Private Image Registry
+
+This tutorial provides to way to utilize bundles from a private image registry in your Entando projects. The steps below use environment variables to pass the Secret for authentication required by private registries.
+
+**New or Existing Bundle**
+* If you have an existing bundle in a private image registry, start at [Step 1: Create the Registry Credentials](#step-1-create-the-registry-credentials).
```
nshaw, line 10: I don't understand lines 10 and 11. Either way you have to do Step 1. What's the distinction?

vuepress/docs/next/tutorials/curate/bundle-private-images.md
```diff
@@ -0,0 +1,85 @@
...
+spec:
+     environmentVariables:
+       - name: ENTANDO_CONTAINER_REGISTRY_CREDENTIALS
+         valueFrom:
+            secretKeyRef:
+               name: container-registry-secret
+               key: registryCredentials
+
+```
+Note: Due to an open issue with Entando 7.1.1, the Entando Component Manager (ECM) requires modification. Scale the deployment down to 0, add the following spec to the ECM deployment, setting the HOME variable, and then restart the cluster.
```
nshaw, line 69: 1) I don't think it should be called ECM in this case. In k8s it's literally YOUR-APPNAME-cm-deployment, e.q. quickstart-cm-deployment` so using 'ecm' will be confusing. 
2) I think it should be numbered as its own step. It's unrelated to #2 so seems odd to mix it in here. When we remove because it's fixed, we can just drop the item. I know I reversed myself here.

PR #620, opened 2022-11-09
ENDOC-606 minor changes

jyunmitch:


Lyd1aCla1r3 (approved):

vuepress/docs/next/docs/getting-started/entando-cli.md
```diff
@@ -28,9 +28,9 @@ The basic requirements for the CLI vary with use case, as shown in the table bel
 
 ## Install the CLI
 Install the current official release of the CLI:
-``` bash
-curl -L https://get.entando.org/cli | bash
-```
+
+<EntandoCode> curl -sL "https://get.entando.org" | ENTANDO_RELEASE="{{ $site.themeConfig.entando.fixpack.v71 }}" bash </EntandoCode>
```
nshaw, line 32: This isn't the same command. Here we want to just install the CLI, not Entando itself. It needs to include /cli on the end. Can you check with Walter on the approved version? In the scratch pad we have `bash <(curl -L "https://get.entando.org/cli") --update --release="v7.1.1" ` but I don't know what the shortest version should be.

nshaw (changes requested):

vuepress/docs/next/docs/getting-started/entando-cli.md
jyunmitch: I checked with Walter and it was the same format as you wrote here & scratchpad. Did not look carefully enough

nshaw (approved):

PR #630, opened 2022-11-22
ENDOC-607 add getting started to left nav in Tutorial Get Started

jyunmitch:
Had to copy the files over to /tutorials/getting-started/ because the left navigation bar does not allow links from tutorials to docs, as far as I can see


vuepress/docs/next/tutorials/getting-started/README.md
```diff
@@ -0,0 +1,372 @@
+---
```
nshaw, line 1: I think this duplicates the full page and assets which will make maintenance harder. I have another idea I'll send over on a new PR.

nshaw (changes requested):

PR #636, opened 2022-12-01
ENDOC-622 add note for run command

jyunmitch:


vuepress/docs/next/tutorials/create/ms/add-api-claim.md
```diff
@@ -61,6 +61,8 @@ ent bundle run simple-node-ms
 ```
 * Access <http://localhost:8081/api/hello> to see a `Hello World!` response with a timestamp. 
 * Access <http://localhost:8081/api/health> to see a standard `{"status:"UP"}` response.
+
+The `run` command is for local testing and uses port 8081 for microservices as a rule.
```
Lyd1aCla1r3, line 65: "uses as a rule" is a little awkward here -> "by default"; a rule implies the port cannot be changed without breaking something

jyunmitch: @nshaw, this is how I understood it, that it would break without a workaround using the params. maybe that info should be included here.

nshaw: Not sure of the language but what we want to call out is that in production the microservice should/must run on port 8081. The run command can be used to modify the port when running it locally so you can run multiple microservices locally side by side without collision.

nshaw (approved):

PR #639, opened 2022-12-02
ENDOC-568 Rewrite ECG page for v5

Lyd1aCla1r3:


vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
-  * Ensure code follows standards and practices of your teams including code formatting, testing tools, static analysis tools, package structures, and other rules specific to your organization
-  * Enable teams to quickly recognize and adopt code from system integrators and other third parties because the patterns and standards are encoded in the starting point of your applications
+- Begin with running software: By automating the creation of microservices and micro frontends, the user is able to bypass the manual development required to build the default configuration of an Entando project.
+- Customize as desired: The generated code is ready to be developed on and modified to enable features and capabilities.
+- Avoid a learning curve: The Entando Blueprint uses standard tools, technologies, and patterns that can be consumed by developers who are not familiar with the underlying technology.
+
+### Standardize Technology 
+
+- Guarantee technology compliance: The Component Generator encodes business standards for networking, monitoring, build tools, and distribution into the Entando Blueprint to ensure that business requirements are satisfied.
```
jyunmitch, line 20: I really like the format with the important point out front separated by a colon.  Just wondering if 'guarantee' might be a wee strong

vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
 
 1. Create components
-   - The Component Generator will create a Spring Boot microservice and optionally will create React micro frontends for entities that you add to your project
-   - The microservice and generated micro frontends are runnable out of the box and preconfigured to connect to [Entando Identity Management](../consume/identity-management.md)
-   - The code is meant to be edited and enhanced by developers. JHipster provides interactive updates and merges if you need to re-run the generator or add new functionality via JHipster
-   - The microservice includes liquibase and integrated data mocking if you want to generate data in your API calls
+   - The Component Generator creates a Spring Boot microservice and, optionally, React micro frontends for entities that you add to your project.
+   - The microservice and generated micro frontends are out-of-the-box runnable and preconfigured to connect to the [Entando Identity Management System](../consume/identity-management-system.md).
```
jyunmitch, line 30: Isn't the microservice also generated? 

can't help it but never liked the word runnable, though it's totally correct. Here though, runnable out of the box sounds better to me. It also matches the flow of the next set of verb and object
are runnable out of the box
are preconfigured to connect

vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
+   - To integrate user edits and enhancements to the code, JHipster offers interactive updates and merges to re-run the modified generator. Alternatively, the user can add new functionality via JHipster.
```
jyunmitch, line 31: can drop user to make first sentence flow nicer, especially since it's repeated in next sentence. not a big deal though

but the generator doesn't get modified, it's just generating with the new updates you give it

vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
+   - The microservice includes Liquibase and integrated mocking to offer data generation in API calls.
```
jyunmitch, line 32: would keep 'data' in place, unless integrated mocking is a 'thing'. looking it up, seems like mocking can apply to different services, and here it's specifically about data.

vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
-  * Code is meant to be developed on and changed by developers. Ready for customization.
-  * Using standard tools, technologies, and patterns that can be consumed by developers who are not familiar with the underlying blueprint
+### Accelerate Development
 
-* Standardize microservice creation and technology choices
-  * Encode business standards for networking, monitoring, build tools, and distribution into a blueprint to ensure that developers are starting from a technical point that matches your teams requirements
-  * Ensure code follows standards and practices of your teams including code formatting, testing tools, static analysis tools, package structures, and other rules specific to your organization
-  * Enable teams to quickly recognize and adopt code from system integrators and other third parties because the patterns and standards are encoded in the starting point of your applications
+- Begin with running software: By automating the creation of microservices and micro frontends, the user is able to bypass the manual development required to build the default configuration of an Entando project.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: i like user here; nothing is lost by broadening the role and "development" already appears in the sentence

vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
+- Customize as desired: The generated code is ready to be developed on and modified to enable features and capabilities.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: agree. if we hadn't inherited it i wouldn't have thought to use the phrase, but it actually is the most accurate and meaningful way to say this, because it's colloquial engineering-speak. developers develop. they either develop something or continue developing something, which is to develop on top of it. that being said, nothing is lost after a little tweaking... "developed on" -> "further developed"

vuepress/docs/next/docs/create/component-gen-overview.md
```diff
@@ -1,42 +1,44 @@
...
+- Avoid a learning curve: The Entando Blueprint uses standard tools, technologies, and patterns that can be consumed by developers who are not familiar with the underlying technology.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: LOL nothing wrong with stating a fact!

vuepress/docs/next/docs/create/component-gen-overview.md
Lyd1aCla1r3: normally i would agree, but the outcome is a sure thing by definition, e.g. if component generator is certified as compliant to industry networking standards, the business requirement for that is satisfied. having a standards compliance background i am sure of this ;)

vuepress/docs/next/docs/create/component-gen-overview.md
Lyd1aCla1r3: i think there's a subtlety here; the MS is generated for sure but the MFE(s) are optional, so it's almost like this is an abbreviation for "the MFEs that you chose to be generated." i'm fine with (optional) replacing it. 

i hate runnable but like the rework. the flow is improved with "the MFEs are 1 and 2" instead of "the MFEs are 1 and are 2." the second "are" can be removed and the previous order can still exist, but "out of the box" is much stronger as an adjective, leading the subject, than an afterthought. here there is only one object and 2 verbs.

vuepress/docs/next/docs/create/component-gen-overview.md
Lyd1aCla1r3: the first instance of user needs to stay to make who is performing the changes explicit. 

the whole point of this bullet is that you can modify the generator, then get all your changes workable to run your new version.

vuepress/docs/next/docs/create/component-gen-overview.md
Lyd1aCla1r3: i don't think any value is lost with removing "data"; the sentence makes it clear that the thing you're handling is data. that being said, no harm in being too clear...

vuepress/docs/next/docs/create/component-gen-overview.md
jyunmitch: disagree on the second point but not at all a big deal

vuepress/docs/next/docs/create/component-gen-overview.md
jyunmitch: @nshaw, is it the blueprint or generator that get customized when using ECG

nshaw: @jyunmitch neither, it's the generated code that gets updated.  The blueprint/generator are the starting points and if you re-run (say by adding an entity or modifying the JDL), it's only the project code that gets modified, not the blueprint/generator itself.

Lyd1aCla1r3: @jyunmitch i think the blueprint and generator are the same thing, right? i had been thinking of the generator as bigger than the blueprint, like implementing the blueprint, but that may not be an accurate distinction. the use of the two terms is a bit confusing tho. wrt the generator itself, the point of the bullet is that you make it your own through modifications. i'll just take out "modified" to not imply that you're tweaking the generator itself.

jyunmitch (approved):

vuepress/docs/next/docs/create/component-gen-overview.md
jyunmitch: Ok, I think the confusion is partly because there are two generators, kind of, the ECR and the JHipster blueprint, and the ECR uses the blueprint?

Lyd1aCla1r3: let's pester @nshaw about this... pretty sure last standup when we asked if they were one and the same i shook my head and you nodded and he said they were synonymous...

PR #641, opened 2022-12-05
ENDOC-626 add note about customizing pack command

jyunmitch:


vuepress/docs/next/docs/curate/bundle-details.md
```diff
@@ -209,18 +209,21 @@ See the [Plugin Environment Variables](../../tutorials/devops/plugin-environment
   ]
 ```
 For more information, go to the [API Management](../getting-started/ent-api.md) page.
+
 ### Command Specification
-|Name|Type|Required|Description|
-|:-|:-|:-|:------------------------|
-|`build`|String|No|Custom build command|
-|`run`|String|No|Custom run command|
+|Name|Type|Required| Default Values| Description|
```
jyunmitch, line 214: All the other tables have 'possible values' but this column is Default but it worked nicer than a note. Unsure if readers will notice this header is different. Can change to note ...

vuepress/docs/next/docs/getting-started/ent-bundle.md
```diff
@@ -104,6 +104,7 @@ See the [Build and Publish a Simple Bundle](../../tutorials/create/pb/publish-si
 - `ent bundle pack`:
    - The artifacts generated for micro frontends and microservices are stored in their respective component folders. File format depends on component type, e.g. a React micro frontend may result in HTML, JavaScript and CSS files.  
    - Once the artifacts are generated, a Docker image for each microservice is built using the Dockerfile located in the microservice's folder. If the Dockerfile is missing, the `pack` command exits with failure.
+   - The `pack` command executes either mvn package or npm run build, depending on the stack. These defaults can be overridden with a [custom command](https://developer.entando.com/v7.1/docs/curate/bundle-details.html#command-specification) section in the bundle descriptor `entando.json`.
```
Lyd1aCla1r3, line 107: commands should be backticked here to avoid the reader thinking there's a missing article ("either an min package," etc.) in a command description and to clarify that they're looking at the commands themselves

custom command section sounds awkward; the contents of that section just boil down to commands, and interpreters work with commands on the lowest level. either custom command or custom commands, depending on if singular or plural is appropriate.

vuepress/docs/next/docs/curate/bundle-details.md
Lyd1aCla1r3: "default values" is not accurate unless both values are used. if it's one or the other it should be "possible values"

vuepress/docs/next/docs/getting-started/ent-bundle.md
jyunmitch: commands backticked

it's a section in the entando.json and it can have multiple commands for the different commands

vuepress/docs/next/docs/curate/bundle-details.md
Lyd1aCla1r3: i guess i don't understand; a default value is what is used without user modification. if that's the case here then it's correct, but if only one of the two is used depending on the stack, then having "default value" as the header with a note to that effect is appropriate

jyunmitch: I was unsure which was more clear, since it doesn't follow the same format as all the other tables on the page. Since it created confusion for you anyways, think it makes more sense to move it below.

PR #648, opened 2022-12-08
Note to connect CLI to an instance before deploy

fuhsaz:
This could perhaps be re-worded, but the gist is that my co-workers and I were running into issues with the `ent bundle deploy` command because we hadn't seen anything about attaching the Entando VM while following the other setup tutorials. Running `ent attach-vm entando` at the suggestion of @nshaw fixed the problem for us, so I thought this note might save future developers a bit of searching.


Lyd1aCla1r3:
hi @fuhsaz, i suggest we modify the language of the prerequisites and linked page to reflect this clearly, before the tutorial is followed.

fuhsaz:
I agree, that's a better place for this. How does this look?

Lyd1aCla1r3:
i think we can combine bullets 2 and 5 and add that command to the linked profile page. if you don't mind, i can create a tracked PR that follows our current processes and style guide, then add you as a reviewer

fuhsaz:
I don't mind at all! Thank you!

Lyd1aCla1r3:
perfect :)

PR #672, opened 2023-01-30
ENDOC-649 left-nav updates

jyunmitch:
Should this be pushed to 7.1?


Lyd1aCla1r3 (approved):

Lyd1aCla1r3:
approved pending conflict resolution

vuepress/docs/.vuepress/next.js
```diff
@@ -424,6 +429,16 @@ module.exports = {
                     path  + 'compose/creating-protected-resources.md'
                 ]
             },
+            {
+                title: 'Consume',
```
nshaw, line 433: Feels a little odd to have Consume Applications under Docs, but just plain Consume under tutorials. Maybe use the same language for both? Consume is the C that seems least intuitive to me so maybe Consume an Application is best, unless the parallel text with "Compose an Application" isn't good.

nshaw (commented):
The distinction between Consume and Config is still a bit vague to me which means I think I'll be checking both places. It seems like Consume should be more directly related to the end user experience but it's not a crisp enough guideline for me to know where to check for a doc, even when I know it exists. Matching the Gartner 4Cs is still worth doing so I don't think that's a blocker, just an observation.

PR #678, opened 2023-02-08
ENDOC-653 update Access Entando APIs

jyunmitch:


vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
-The `errors` will always contain code and a message string indicating an
-error condition in the request. The `metadata` section is used for
-paging, sorting, filtering and data that is distinct from the body. The
-body of each response is included in the `payload` section of the
-response and varies according to each API.
+## Find Your Client Secret
+You'll need your client credentials to execute the Entando APIs. 
 
-### Models
+1. Login into your Keycloak Administration Console. To find the Kubernetes credentials, see the [Entando Identity Management System](./identity-management.md) page.
```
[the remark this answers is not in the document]
nshaw: @jyunmitch good catch, no, it's just the admin credentials which can be found in the corresponding k8s secret. "To find the Kubernetes credentials" -> "To find the admin credentials for Keycloak"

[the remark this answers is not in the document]
jyunmitch: @nshaw, was trying to avoid "admin" credentials because that is used below in the Swagger authorization steps below (Access APIs on Swagger #4) for the Keycloak admin user (admin/adminadmin) and I wanted to try to distinguish the two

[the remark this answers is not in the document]
nshaw: @jyunmitch Hmm. "Keycloak admin user" != "admin/adminadmin".  "admin" is the Entando Application admin user.  "entando_keycloak_admin" is the Keycloak admin user. Neither of them are Kubernetes users or credentials.  Not sure that helps.  :(

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
-
-### Configure Keycloak
-
-Configure your Keycloak client in order to support Swagger UI. A quickstart environment has this pre-configured.
 
-1. Login to your Keycloak instance
+2. (Optional) Scale the deployment `spec.replicas` to 0 before updating the deployment. This is necessary if you're using an in-memory database, e.g. the default quickstart configuration, and will prevent database errors that can happen on an immediate restart after the profile is changed. Save the deployment to update. 
 
-2. Access the Administration console
+3. In the deployment YAML, find the env variables section under `spec.template.spec.containers.env[image: entando-de-app]`
```
Lyd1aCla1r3, line 33: either spell out the abbreviation "environment variables" or format "env variables" with backticks, etc., so the reader knows this is the actual name of the section and not a casual instruction

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
-## Setup in local environment
-
-You may prefer to run a local standalone Entando application for some tasks. You'll need Java 11, maven, and Keycloak for authentication. See [these instructions](https://github.com/entando/app-builder/blob/master/with-keycloak.md) to setup a standalone Keycloak.
-
-### Configure Keycloak
-
-Configure your Keycloak client in order to support Swagger UI. A quickstart environment has this pre-configured.
 
-1. Login to your Keycloak instance
+2. (Optional) Scale the deployment `spec.replicas` to 0 before updating the deployment. This is necessary if you're using an in-memory database, e.g. the default quickstart configuration, and will prevent database errors that can happen on an immediate restart after the profile is changed. Save the deployment to update.
```
Lyd1aCla1r3, line 31: first sentence says what you need to do before you update, last sentence says what to do to update - applicability sentence between them makes this disjointed and not a clear instruction. can be combined into a single directive: "(Optional) Scale the deployment `spec.replicas` to 0, then save the deployment to update it." 

second sentence can be simplified a bit: "This is necessary if you're using an in-memory database, e.g. the default quickstart configuration, to prevent database errors triggered by an immediate restart after the profile is changed."

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
+        - name: SPRING_PROFILES_ACTIVE
+          value: default,swagger
+```
 
-The Entando core exposes REST APIs for every action that can be taken in
-the App Builder environment. For example, you can use
-these APIs to create pages, create page templates or to add widgets to
-pages. The APIs can be used to support automation, testing, or
-integrations with external systems.
+5. (Optional) Reset the deployment `spec.replicas` back to 1 if it was scaled to "0" in the previous steps. Save the deployment to update.
```
Lyd1aCla1r3, line 41: "..scaled to 0 **for** the previous **steps**" or "scaled to 0 **in** a previous **step**" (it is not scaled to 0 in each of the previous steps, which is how this reads)

0 or "0"?? in step 2 there are no quotes...

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3, line 50: Log in

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
-body of each response is included in the `payload` section of the
-response and varies according to each API.
+## Find Your Client Secret
+You'll need your client credentials to execute the Entando APIs. 
 
-### Models
+1. Login into your Keycloak Administration Console. To find the Kubernetes credentials, see the [Entando Identity Management System](./identity-management.md) page.
 
-All of the model classes returned by the Entando core are annotated so that the model definition is included in the Swagger documentation. At the bottom of the Swagger page all of the model classes returned by the API endpoints can be found.
+2. From the left nav, go to `Clients`
```
Lyd1aCla1r3, line 52: left navigation panel or menu on the left... "left nav" is not clear to everyone and very casual

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-## Tutorial
+3. Select the desired client (e.g. in a quickstart environment this is `quickstart`)
```
Lyd1aCla1r3, line 54: environment, (comma)

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-1. Access your application Swagger UI as discussed above
+4. Click on the `Credentials` tab to retrieve the secret
```
Lyd1aCla1r3, line 56: Secret

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-2. Click on the `Authorize` button in the upper right corner
+## Access the APIs on Swagger
+1.  To see the APIs, go to: 
+``` sh
+http://[YOUR-HOST-NAME]/entando-de-app/api/swagger-ui.html
+```
 
-3. Enter the client id and client secret in the open window and click `Authorize`
+2. Click on the Authorize button in the upper right corner.
```
Lyd1aCla1r3, line 64: notice that there are periods for single-sentence steps here (no periods used for single sentences in above steps)

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-4. If you are redirected to the Entando login page, log in with your credentials (default are `admin`/`adminadmin`)
+3. Enter the client id and secret and click Authorize.
```
Lyd1aCla1r3, line 66: same wrt period

Secret

ID should be capitalized unless it's the property itself, then should be backticked

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-5. You will be redirected to the Swagger UI page, now authenticated
+4. You will be prompted to log in to your Keycloak instance as an admin user if you have not already done so. The defaults are admin/adminadmin. 
 
-6. Use the **Try it out** button on the APIs
+5. You will be redirected to the authenticated Swagger UI page. Select an API to see the methods in the drop-down list and click the `Try it out` button.
```
Lyd1aCla1r3, line 70: **its** methods (not all methods are the same??)

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
+   1. Scroll down to `widget-controller`
 
-    -   Scroll to `widget-controller`
+   2. Select the `GET` method
 
-    -   Select the blue GET row
+   3. Click `Try it out`
 
-    -   Select **Try it out**
+   4. See the results in the window below. You should see a Server response with Code 200 and full response body.
```
Lyd1aCla1r3, line 78: "The results are visible in the window below" or "view the results..." (a bit less aggressive and avoids a repeat of the word "see")

"see a Server response with Code 200 and full response body"... why is server capitalized? if code is capitalized "Code 200" should be in quotes to indicate that's the actual value you see (conversely code 200 implies you'll see just the number 200)

"You should see a Server response with Code 200 and full response body"... i feel like this could be rephrased for clarity and to avoid two "responses"

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-    -   Look at the results in the window. You should see a Server response with Code 200 and full response body.
\ No newline at end of file
+### Next Steps
+* For information, see the tutorial on [adding a REST API](../../tutorials/devops/add-rest-api.md).
+* Learn how to [invoke Entando core APIs](../../tutorials/devops/invoking-api.md).
```
Lyd1aCla1r3, line 82: here is Entando core again! are "entando" and "core" modifiers for "APIs" or is "Entando" modifying "core APIs"? like are we talking about the Entando core or core APIs? if the entando core is a thing we're referencing often we need to better define it in an intro page

Lyd1aCla1r3: no mention of "core" here: https://developer.entando.com/next/docs/getting-started/concepts-overview.html#key-entando-concepts

vuepress/docs/next/docs/consume/entando-apis.md
jyunmitch: not sure why server was capped??
Code 200 is like Error 404, does it need backticks?? In this case, don't think it's necessary
edited

I actually prefer 'see the results..' it's direct and clear. view sounds too much like sightseeing.

vuepress/docs/next/docs/consume/entando-apis.md
jyunmitch: Not touching this one because it is the tutorial name and I think it refers to core in the sense that it's attached to the center of Entando which is the Engine. If we want to change that title, we can get a ticket for that. Core and engine are used interchangeably by the devs apparently

jyunmitch: got rid of both next steps because the REST apis is going to be deleted anyways

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
 ---
 # Accessing Entando APIs
 
-## Overview
 
-Entando includes the Swagger UI in a quickstart environment and is reachable at `/entando-de-app/api/swagger-ui.html`:
+Entando includes the Swagger UI for API access in a quickstart environment. This document presents an overview and how to enable and access the Swagger UI for Entando APIs.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: not sure i understand your comment; neither version of the sentence provides a reference or context for what the quickstart env is... rereading i think i like the original better tho anyway. 

in terms of defining the quickstart env, isn't this the env you configure when you follow our getting started guides? we reference the quickstart env often and it's not explicitly defined anywhere, so i think maybe we should think about doing that. linking to that can be done as necessary when we revisit pages, or in a separate ticket to hit all the mentions

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-    http://[your-host-name]/entando-de-app/api/swagger-ui.html
+## APIs Overview
+The Entando core exposes REST APIs for every action that can be taken in the App Builder. For example, you can use APIs to create pages, page templates or add widgets to pages. The APIs can be used to support automation, testing, or integrations with external systems as well.
 
-### Enable or disable the Swagger UI in a running container
+## API Structure
+All the APIs share a common top level structure. Each response will contain a top level entry for errors, metadata, and payload.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: i don't understand this comment. can you rephrase to clarify?

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-The Swagger UI can be enabled or disabled in a running container by modifying the SPRING_PROFILES_ACTIVE environment variable for the entando-de-app container. 
+The errors will always contain code and a message string indicating an error condition in the request. The metadata section is used for paging, sorting, filtering and data that is distinct from the body. The body of each response is included in the payload section of the response and varies according to each API.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: aha. the repetition threw me. if this level of detail is necessary it's probably worth clarifying a bit

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-1. Edit the deployment. The name may be different outside of a quickstart environment. Note: Use the [ent CLI](../getting-started/entando-cli.md) to send commands to Kubernetes from the host machine.
-```
-kubectl -n entando edit deployment/quickstart-server-deployment
-```
+## Models
+All of the model classes returned by the Entando core are annotated so that the model definition is included in the Swagger documentation. At the bottom of the Swagger page all of the model classes returned by the API endpoints can be found.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: another aha. your explanation connects the dots... if the first sentence can be modified to include the gist of that it would be worthwhile. like "All of the model classes for the data returned by the App Engine are annotated, with the annotations defined in the Swagger documentation." this is how i read your comment, which is a bit different from - and more clear than - the 1st sentence as-is

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,80 @@ sidebarDepth: 2
...
 
-2. (Optional) Scale the deployment `spec.replicas` to 0 before updating the deployment. This is necessary if you're using an in-memory database, e.g. the default quickstart configuration, and will prevent database errors that can happen on an immediate restart after the profile is changed. Save the deployment to apply the change. 
+## Enable the Swagger UI
 
-3. Find the entando-de-app env variables section under `spec.template.spec.containers.env[image: entando-de-app]`
+The Swagger UI can be enabled or disabled in a running Entando instance by modifying the `SPRING_PROFILES_ACTIVE` environment variable for the `entando-de-app` container.
```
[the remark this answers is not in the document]
Lyd1aCla1r3: that's fair; my concern is that someone will gloss over the importance of "running Entando instance" if they're not familiar with this, like that's an important detail. it *should* be obvious but i could see myself trying to follow the tutorial locally or with my vm stopped if i didn't have experience or know to ask questions

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3: no no no, sorry - i mean do the same methods apply to all the APIs or do different APIs support different methods? if the same always apply then "the methods" is the way to go because there is only one set and it's used for every API - the specificity works. but if different APIs have different methods then "the" APIs is imprecise, but "its APIs" is specific

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3: i don't think backticks, only quotes because it's a name

we should def go for the sightseeing angle. maybe we could start selling merch ;)

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3: or we could make a ticket for defining core somewhere in the docs... it appears here and there but is never defined

vuepress/docs/next/docs/consume/entando-apis.md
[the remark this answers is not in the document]
jyunmitch: anyone looking at rest APIs knows what bodys are

vuepress/docs/next/docs/consume/entando-apis.md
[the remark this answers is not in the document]
jyunmitch: it's been changed already

vuepress/docs/next/docs/consume/entando-apis.md
jyunmitch: It's not a name, it's a resultant code and in HTTP protocols, it's a common response that really doesn't need special citing

vuepress/docs/next/docs/consume/entando-apis.md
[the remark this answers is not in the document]
jyunmitch: In this version: 'For API access in a quickstart environment,...' the quickstart enve has no reference point, whereas starting with Entando gives it the reference that it's an Entando environment

vuepress/docs/next/docs/consume/entando-apis.md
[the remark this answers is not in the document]
Lyd1aCla1r3: as an example, "all the kids eat cake" vs "all of the kids eat cake"; neither clarifies whether they're eating the same cake (sharing something together) or their own cakes (have eating cake in common). not worth pursuing

vuepress/docs/next/docs/consume/entando-apis.md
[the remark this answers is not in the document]
Lyd1aCla1r3: not what i meant at all

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3: not sure what the GET method has to do with this? pretty sure most everyone would use "its"
Screen Shot 2023-02-17 at 8 50 08 AM

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3: it's the name of a code. i'm fine with capitalized and no quotes, but we have quoted or otherwise formatted codes elsewhere in the documentation so this is more about consistency. should we say going forward to not format stuff like this? i actually think we should because it gets really busy with backticks/quotes and like you said, it's not necessary

vuepress/docs/next/docs/consume/entando-apis.md
Lyd1aCla1r3: great. that clears it up

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -63,20 +65,16 @@ http://[YOUR-HOST-NAME]/entando-de-app/api/swagger-ui.html
 
 2. Click on the Authorize button in the upper right corner.
 
-3. Enter the client id and secret and click Authorize.
+3. Enter the Client ID and Secret from above. Click Authorize.
```
Lyd1aCla1r3, line 68: `Client Id` like in step 4??

vuepress/docs/v7.1/docs/consume/entando-apis.md
```diff
@@ -3,139 +3,78 @@ sidebarDepth: 2
...
+1.  To see the APIs, go to: 
+``` sh
+http://[YOUR-HOST-NAME]/entando-de-app/api/swagger-ui.html
+```
 
-4. If you are redirected to the Entando login page, log in with your credentials (default are `admin`/`adminadmin`)
+2. Click on the Authorize button in the upper right corner.
 
-5. You will be redirected to the Swagger UI page, now authenticated
+3. Enter the Client ID and Secret from above. Click Authorize.
```
Lyd1aCla1r3, line 68: same comment

vuepress/docs/next/docs/consume/entando-apis.md
```diff
@@ -4,56 +4,58 @@ sidebarDepth: 2
 # Accessing Entando APIs
 
 
-Entando includes the Swagger UI for API access in a quickstart environment. This document presents an overview and how to enable and access the Swagger UI for Entando APIs.
+Entando includes the Swagger UI for API access in a quickstart environment. This document presents an overview and instructions on how to enable and access the Swagger UI.
 
 ## APIs Overview
-The Entando core exposes REST APIs for every action that can be taken in the App Builder. For example, you can use APIs to create pages, page templates or add widgets to pages. The APIs can be used to support automation, testing, or integrations with external systems as well.
+The Entando App Engine uses REST APIs to enact all the functionality inside the App Builder. For example, APIs are used to add widgets to a page or create components like pages and page templates. APIs can also be used to support automation, testing, and integration with external systems.
```
Lyd1aCla1r3, line 10: could enact -> execute? seems to be more precise and the usual language?

Lyd1aCla1r3 (approved):
i really like it now :) 2 small comments but i'll go ahead and approve now

vuepress/docs/next/docs/consume/entando-apis.md
jyunmitch: trying to minimize backticking... here backticked client id is the same as the literal words client ID the term. I don't think there would be any confusion.

vuepress/docs/next/docs/consume/entando-apis.md
jyunmitch: I think enact is fine here, especially as APIs don't really execute things as they just move information back and forth

PR #685, opened 2023-02-22
ENDOC-585 REST API update & custom stack bit

jyunmitch:


vuepress/docs/next/docs/curate/bundle-details.md
```diff
@@ -176,12 +176,17 @@ See the [Plugin Environment Variables](../../tutorials/devops/plugin-environment
...
+      "stack": "custom",
       "type": "widget",
       "publicFolder": "public",
       "titles": { "en": "My MFE2 Title", "it": "Il Mio Titolo MFE2" },
       "group": "free",
-      "commands": { "build": "custom-command" }
+      "commands": {
+           "build": "npm run build",
+           "run": "npm install -g http-server",
+            "pack": "docker build ..."
```
jyunmitch, line 187: @nshaw, can you correct this? I tried to add an example and could not find an equivalent for the 3 commands for build, run, and pack. Maybe this is unnecessary.. LMK

nshaw: @jyunmitch I might use the echo examples from https://entando.myjetbrains.com/youtrack/issue/ENG-4175. That helps communicate the 3 commands that can/should be customized but we don't have to determine realistic commands for an unknown stack.

vuepress/docs/next/docs/curate/bundle-details.md
```diff
@@ -133,13 +133,13 @@ The following is a list of specifications for the bundle descriptor and its comp
     }
   ]
 ```
-::: tip  
- Entando uses the `healthCheckPath` to monitor the health of the microservice. A plugin in an Entando Bundle can use any technology, as long as it provides a health check service configured via the `healthCheckPath`. This path must be specified in the descriptor file and return an HTTP 200 or success status. This can be implemented by a Java service included with the Entando Blueprint in the Spring Boot application. You can also [use a Node.js service as shown here](https://github.com/entando-samples/ent-project-template-node-ms/blob/main/src/main/node/controller/health-controller.js). 
-:::
+#### Microservices Details
+ - It is recommended that REST APIs be added to microservices on Entando, following REST frameworks.
```
Lyd1aCla1r3, line 137: this sentence needs some clarification; it reads like REST APIs should be added to MSs no matter what, not that if a dev wants to use a REST API they way they should go about that is by adding it to an MS. also, "to MS on Entando" reads like REST APIs should be added after-the-fact (as opposed to whenever - during or after MS creation)

following REST frameworks as in according to rules and guidelines imposed by REST frameworks? "according to REST frameworks" is more precise and has only one meaning, whereas "following" can read as either "in line with" or "after"

"it is recommended"... who is recommending this and why? that info is a nice-to-have

suggest something like: "To make use of REST APIs on Entando, add them to microservices in accordance with REST frameworks."

vuepress/docs/next/docs/getting-started/ent-bundle.md
```diff
@@ -61,6 +61,17 @@ See the [Build and Publish a Simple Bundle](../../tutorials/create/pb/publish-si
 
 - `--hub-url`: Use this option to specify a custom Entando Hub, else ent defaults to the Entando Cloud Hub
 
+### Bundle Micro Frontend and Microservice
+| Command| Descriptions
+|:--|:--
+|`ent bundle mfe add` --stack [stack-type] [name] | Add a default React or Angular micro frontend |
+|`ent bundle mfe add` --stack custom [name] | Add custom stack micro frontend |
```
Lyd1aCla1r3, line 68: Add a

vuepress/docs/next/docs/getting-started/ent-bundle.md
```diff
@@ -61,6 +61,17 @@ See the [Build and Publish a Simple Bundle](../../tutorials/create/pb/publish-si
...
+|`ent bundle ms add --stack [stack-type] [name]` | Add a default Spring Boot or node microservice |     
+|`ent bundle ms add` --stack custom [name]` | Add a custom stack microservice |
+
+#### Command Details
+- `ent bundle mfe add --stack custom [name]` & `ent bundle ms add --stack custom [name]`: When a custom stack is used for any MFE or microservice, custom commands for the build, run, and pack functions are required under the Command spec. In addition, a version number must be specified for each. Because Entando can only parse information from a pom.xml or package.json, the custom commands are required to manage the bundle and the version number is required to track Docker images. For an example, see the [Bundle Details](../curate/bundle-details.md#micro-frontends-sample-code) page.
```
Lyd1aCla1r3, line 73: 1st: "...Command spec" -> should indicate where this is found, e.g. "...Command spec of the `entando.json`."

2nd: "...for each." leaves the reader hanging; "for each command" would be complete.

3rd: "...manage the bundle and the version number..." -> comma after "and"; without the comma the reader is expecting that the custom commands are required to manage the bundle and something else, not to manage the bundle *pause* and something else is required to....

vuepress/docs/next/tutorials/create/ms/add-api-claim.md
```diff
@@ -50,6 +50,8 @@ app.listen(port, function () {
     console.log('Listening to Port ' + port);
 }); 
 ```
+>REST APIs should be added to microservices on Entando, following REST frameworks.
```
Lyd1aCla1r3, line 53: same as above

vuepress/docs/next/tutorials/create/ms/add-api-claim.md
```diff
@@ -50,7 +50,7 @@ app.listen(port, function () {
     console.log('Listening to Port ' + port);
 }); 
 ```
->REST APIs should be added to microservices on Entando, following REST frameworks.
+>Entando recommends adding REST APIs to microservices.
```
jyunmitch, line 53: @nshaw, needed to add note about adding REST APIs to MSs but it looks out of place here in this tutorial. What do you think? It was hard to find an appropriate place for it. Another alternative is the api-management.md but that's for the ent CLI info and that seemed odd too.

vuepress/docs/v7.1/tutorials/create/ms/add-api-claim.md
```diff
@@ -3,6 +3,8 @@ Micro frontends typically require data or logic provided by microservices. The c
 
 :::tip Note
 API claims in Entando can be used for microservices within the same bundle or for microservices within another bundle in the same namespace. This tutorial demonstrates the first case. 
+
+Entando recommends creating REST APIs in microservices instead of extending the App Engine as previously done.
```
Lyd1aCla1r3, line 7: super minor - can remove "as previously done" because why draw attention to a method that is no longer recommended/supported... but keeping "instead of extending the app engine" is good - someone might decide that's another way to do it that's just as acceptable or easier, etc

Lyd1aCla1r3 (approved):

vuepress/docs/v7.1/tutorials/create/ms/add-api-claim.md
jyunmitch: I agree with you about not wanting to draw attention to it and I waffled both ways. In the end, I think it gives it context that is helpful. I do like how it sounds without 'as previously done' but ending it like that seemed to infer they are both ok options but one is just more recommended. This way, people see it is an evolution

PR #724, opened 2023-04-27
Endoc 673- MT CDS tutorial

jyunmitch:
@avdev4j , use this to test CDS. This has been updated with the help of PJ & Eugenio. The google doc I linked before is outdated.

### DIFF
--- a/vuepress/docs/next/tutorials/consume/mt-cds.md
+++ b/vuepress/docs/next/tutorials/consume/mt-cds.md
@@ -3,45 +3,46 @@ sidebarDepth: 2
 ---
 
 # Content Delivery Server for Multitenancy
-You can create a content delivery server (CDS) using descriptors or a custom script to set up the configuration. Then adapt the Entando App Engine to integrate the service for your multitenant application. This tutorial applies descriptors to build the CDS.
+To create a content delivery server (CDS), use descriptors or a custom script to set up the configuration. A CDS is required for a multitenant application. Then the Entando App Engine deployment is edited to integrate the service. This tutorial describes the steps required to create and apply the CDS descriptors to support a multitenant application.
 
 ## Prerequisites
 * [A working instance of Entando.](../../../docs/getting-started/README.md) with the default Tomcat server image.
 
+* Verify dependencies with the [Entando CLI](../../../docs/getting-started/entando-cli.md#check-the-environment): `ent check-env develop`
+
 ## Create the CDS Descriptors
 Descriptors for Keycloak access, ingress, service/deployment, and persistent volume claim are required for the CDS.
+
+Here is a list of placeholders that will need to be replaced in the descriptor files. Use names that describe your primary tenant and the service it pertains to. 
+| Placeholder | Descriptions 
+|:--|:--
+| YOUR-APPNAME | Your app name, or the name of the entando-de-app deployment
+| YOUR-HOSTNAME | Host name (e.g., k8s-entando.org)
+| YOUR-NAMESPACE | Namespace used for the multitenancy application
+| YOUR-PRIMARY-KC-SECRET | Name of Keycloak secret for the primary tenant
+| YOUR-PRIMARY-CDS-DEPLOYMENT | CDS deployment for the primary tenant
+| YOUR-PRIMARY-PVC | Name of PVC claim of the primary tenant
+| YOUR-PRIMARY-CDS-SERVICE | The CDS service for the primary tenant
+| YOUR-PRIMARY-CDS-DEPLOYMENT | The CDS deployment of the primary tenant
+| YOUR-PRIMARY-CDS-INGRESS | The CDS ingress for the primary tenant
+| YOUR-PRIMARY-TENANT | Name of the primary tenant and primary tenant code 
+| YOUR-PUBLIC-KEY | Keycloak RSA generated string for your realm
 
 ### Keycloak Descriptor YAML
 1. Go to Keycloak admin console and log in
 2. From the left sidebar, `Realm Settings` → `Keys`. Click on `Public Key` for `rsa-generated` provider and copy the content.
 
-<details><summary><b>Alternate Method to find the Public Key</b></summary>
-
-1. Retrieve you ingress:
-```
-kubectl get ing | grep default-sso | awk '{print $3} 
-```
-The result of this query will be YOUR-INGRESS
-
-2. Retrieve your public key:
-
-```
-curl "http://YOUR-INGRESS/auth/realms/YOUR-NAMESPACE" | jq '.public_key'
-```
-
-</details>
-
-3. Add the public key to your descriptor file, following the example shown here: 
+3. Create a descriptor using this example. Replace `YOUR-PUBLIC-KEY` with the result from the previous step, following the `---BEGIN ... END---\n` format shown below:
 
 ``` yaml
 apiVersion: v1
 kind: Secret
 metadata:
-  name: YOUR-KC-SECRET
+  name: YOUR-PRIMARY-KC-SECRET
   namespace: YOUR-NAMESPACE
 type: Opaque
 stringData:
-  KC_PUBLIC_KEY: "YOUR-PUBLIC-KEY"
+  KC_PUBLIC_KEY: "-----BEGIN PUBLIC KEY-----\nYOUR-PUBLIC-KEY\n-----END PUBLIC KEY-----\n"
 ```
 
 ### Persistent Volume Claim Descriptor
@@ -52,30 +53,30 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
-    deployment: YOUR-CDS-DEPLOYMENT
-  name: YOUR-PVC-CLAIM
+    deployment: YOUR-PRIMARY-CDS-DEPLOYMENT
+  name: YOUR-PRIMARY-PVC
   namespace: YOUR-NAMESPACE
 spec:
   accessModes:
     - ReadWriteOnce
   resources:
     limits:
-      storage: PVC-SIZE
+      storage: 1Gi
     requests:
-      storage: PVC-SIZE
+      storage: 1Gi
 ```
 
 ### Service and Deployment Descriptor
-Create the service and deployment descriptor YAML with the specifications required for your environment in a K3s implementation.
+Create the service and deployment descriptor YAML with the specifications required for your environment in a K3s implementation. For convenience, the service and deployment has been combined into a single descriptor `YOUR-PRIMARY-CDS-DEPLOYMENT.yaml`, but this is a matter of preference.
 
 ``` yaml
 apiVersion: v1
 kind: Service
 metadata:
-  name: YOUR-CDS-DEPLOYMENT
+  name: YOUR-PRIMARY-CDS-SERVICE
   namespace: YOUR-NAMESPACE
   labels:
-    app: YOUR-CDS-DEPLOYMENT
+    app: YOUR-PRIMARY-CDS-SERVICE
 spec:
   ports:
     - port: 8080
@@ -87,25 +88,33 @@ spec:
       protocol: TCP
       targetPort: 8081
   selector:
-    app: YOUR-CDS-DEPLOYMENT
+    app: YOUR-PRIMARY-CDS-DEPLOYMENT
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: YOUR-CDS-DEPLOYMENT
+  name: YOUR-PRIMARY-CDS-DEPLOYMENT
   namespace: YOUR-NAMESPACE
   labels:
-    app: YOUR-CDS-DEPLOYMENT
+    app: YOUR-PRIMARY-CDS-DEPLOYMENT
 spec:
   selector:
     matchLabels:
-      app: YOUR-CDS-DEPLOYMENT
+      app: YOUR-PRIMARY-CDS-DEPLOYMENT
   template:
     metadata:
       labels:
-        app: YOUR-CDS-DEPLOYMENT
+        app: YOUR-PRIMARY-CDS-DEPLOYMENT
     spec:
-      containers:
+      initContainers:
+	        - name: init-cds-data
+	           image: busybox
+	           imagePullPolicy: IfNotPresent
+	           command: ['sh', '-c', 'wget --no-check-certificate https://raw.github.com/entando-ps/cds/pr-it-317/entando-data/entando720.tar.gz; tar -xf entando720.tar.gz; rm entando720.tar.gz ']
+	           volumeMounts:
+	             - mountPath: /entando-data
+	               name: cds-data-volume
+	    containers:
         - readinessProbe:
             httpGet:
               port: 8081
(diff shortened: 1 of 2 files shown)


vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+          - backend:
+              service:
+                name: YOUR-APP-NAME-cds-YOUR-TENANT-CODE-service
+                port:
+                  number: 8081
+            pathType: Prefix
+            path: /$TN
+          - backend:
+              service:
+                name: $APP-cds-$TN-service
```
eugeniosant, line 198: Should we standardize naming conventions (like i see in this attribute) in each yaml or not?

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+## Create the Assets Archive 
+Create an archive of the resources and assets for your application.  Within it, the resources folder must be named `public` and sensitive information should be stored in the `private` folder.
+
+## APIs
+Create API requests to include these fields:
+ * `CDS_PATH` : # the base path of the new CDS
+ * `KC_URI` : # the url of token service of the tenant1 realm
+ * `KC_CLIENT_ID` : # the clientId of the reserved client used by the entando-de-app in tenant1 realm
+ **what is the reserved client???**
+ * `KC_CLIENT_SECRET` : # the secret of that client
```
eugeniosant, line 253: these are the parameters used by the Postman collection created for loading the cds, and without that collection these parameters have no sense (and "tenant1" was indicated). So either you propose that collection or you have to document all the API calls that need to be made for uploading the archive.

eugeniosant (changes requested):

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Keycloak Descriptor YAML
+1. Go to Keycloak admin console and log in
+2. From the left sidebar, `Realm Settings` → `Keys`. Click on `Public Key` for `rsa-generated` provider and copy the content.
+
+<details><summary><b>Alternate Method to find the Public Key</b></summary>
+
+1. Retrieve you ingress:
+```
+kubectl get ing | grep default-sso | awk '{print $3}
```
avdev4j, line 22:
```suggestion
kubectl get ing | grep default-sso | awk '{print $3}' 
```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+1. Retrieve you ingress:
+```
+kubectl get ing | grep default-sso | awk '{print $3} 
+```
+The result of this query will be YOUR-INGRESS
+
+2. Retrieve your public key:
+
+```
+curl "http://YOUR-INGRESS/auth/realms/YOUR-NAMESPACE" | jq '.public_key'
```
avdev4j, line 29: In my case, the realm name define here as  "YOUR-NAMESPACE" was "entando" not my namespace value. I don't know if it is always as is, but we should have the right realm value.

avdev4j: btw, using our env http has to be https instead

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+    - ReadWriteOnce
+  resources:
+    limits:
+      storage: PVC-SIZE
+    requests:
+      storage: PVC-SIZE
+```
+
+### Service and Deployment Descriptor
+Create the service and deployment descriptor YAML with the specifications required for your environment in a K3s implementation.
```
avdev4j, line 69: is it only working in a k3s env?

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+  replicas: 1
+```
+
+### Ingress Descriptor
+``` yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: YOUR-INGRESS-NAME
+  namespace: YOUR-NAMSPACE
```
avdev4j, line 173:
```suggestion
  namespace: YOUR-NAMESPACE
```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+#  entando.assume.external.https.provider: "true"
+#  entando.k8s.operator.impose.limits: "true"
+```
+
+</details>
+
+## Apply the Descriptors
+Install the CDS descriptors in the order listed below.
+``` bash
+kubectl apply -f YOUR-CDS-PVC.yaml
```
avdev4j, line 237:
```suggestion
kubectl apply -f YOUR-CDS-PVC.yaml -n YOUR-NAMESPACE
```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+kubectl apply -f YOUR-CDS-KEYCLOAK-SECRET.yaml
```
avdev4j, line 238:
```suggestion
kubectl apply -f YOUR-CDS-KEYCLOAK-SECRET.yaml -n YOUR-NAMESPACE
```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+kubectl apply -f YOUR-CDS-DEPLOYMENT.yaml
```
avdev4j, line 239:
```suggestion
kubectl apply -f YOUR-CDS-DEPLOYMENT.yaml -n YOUR-NAMESPACE
```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+kubectl apply -f YOUR-CDS-INGRESS.yaml
```
avdev4j, line 240:
```suggestion
kubectl apply -f YOUR-CDS-INGRESS.yaml -n YOUR-NAMESPACE
```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      proxy_set_header X-Forwarded-For \$proxy_add_x_forwarded_for;
+      proxy_set_header X-Scheme \$scheme;
+      proxy_set_header X-Forwarded-Proto \$scheme;
+      add_header Content-Security-Policy upgrade-insecure-requests;
+    nginx.ingress.kubernetes.io/proxy-body-size: "150m" # edit according to the file size you require
+
+    nginx.org/client-max-body-size: "150m" # edit according to the file size you require
+spec:
+  $INGRESS_CLASS_FIELD
```
avdev4j, line 184: I guess this should be not here, it might be `ingressClassName: nginx` instead.

avdev4j: It causes an error
`error: error parsing cds-ingress-descriptor.yaml: error converting YAML to JSON: yaml: line 17: mapping values are not allowed in this context`

vuepress/docs/next/tutorials/consume/mt-cds.md
avdev4j: @eugeniosant I assume you are talking about strings to be replaced, if yes I guess we should standardize them. At least to make it clear they have to be replaced

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+## Prerequisites
+* [A working instance of Entando.](../../../docs/getting-started/README.md) with the default Tomcat server image.
+
+## Create the CDS Descriptors
+Descriptors for Keycloak access, ingress, service/deployment, and persistent volume claim are required for the CDS.
+
+### Keycloak Descriptor YAML
+1. Go to Keycloak admin console and log in
+2. From the left sidebar, `Realm Settings` → `Keys`. Click on `Public Key` for `rsa-generated` provider and copy the content.
+
```
nshaw, line 17:
```suggestion
3.  In the steps below, use the following format and replace YOUR-PUBLIC-KEY with the result of the previous step:

   KEYCLOAK_PUBLIC_KEY="-----BEGIN PUBLIC KEY-----\nYOUR-PUBLIC-KEY\n-----END PUBLIC KEY-----\n"

```

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+<details><summary><b>Alternate Method to find the Public Key</b></summary>
```
nshaw, line 18: Let's remove the alternate steps so we don't have to maintain both.

vuepress/docs/next/tutorials/consume/mt-cds.md
avdev4j: @eugeniosant Is that section really needed here? As far as I understand we can remove it because it requires more explanation than what we have (configuration Postman for example).

If my understanding is good, after having updated the entando-de-app I would be able to upload files to the CDS through the app-builder, right?

If we have to keep that section to upload the assets, we have to improve it to make it clear what the user is suppose to do.

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+## Configure the Entando App Engine
+1. Scale to 0 the de-app deployment
+2. Open `entando-de-app` and add these environment variables using the :
+``` yaml
+spec:
+   env:
+     - name: CDS_ENABLED
+       value: "true"
+     - name: CDS_PUBLIC_URL
+       value: http://cds.YOUR-HOSTNAME/primary
```
avdev4j, line 264: @eugeniosant is the path here have to be changed according to the ingress value defined before? If yes, it means we should use a placeholder here.

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+     - name: CDS_PRIVATE_URL
+       value: http://cds:8080
+     - name: CDS_PATH
+       value: /api/v1
+```
+3. Delete the following two volume specifications related to the quickstart deployment:
```
avdev4j, line 270: Is this still have to be done even if we are not using the quickstart installation method? I think so but the volume names would not follow that pattern `quickstart...` we should use placeholders too.

avdev4j (changes requested):

avdev4j (changes requested):

avdev4j:
## A summary of the current tests

- [X] The CDS resources can be installed (with modification in the sample files)
- [X] The resources run and know each other (ingress, service, deployment, secret, and pods)
- [X] The CDS API is reachable outside (through the ingress), and we can upload a file
- [X] The entando-de-app works using the CDS
- [ ] The app-builder is running well using the CDS 

### Minor
- Typo fixing
    - Ensure placeholders use the correct pattern
    - Ensure placeholders have the correct value
    - Minor changes in commands
    - Use a suitable placeholder to avoid misunderstanding the usage of that value
    - Streamline placeholders
 
My suggestion: 
because we have a lot of placeholders to replace each is not easy to understand what kind of value we expect, we should define all of them before including the placeholder name, and a definition. To recognize them easily we can wrap them using brackets or equivalent. We can also suggest a pattern for tenants.
 e.g. <YOUR-CDS-DEPLOYMENT>: The name of the deployment resource defines in Kubernetes or to rely on in other descriptors. ex: `cds-deployment-tenant1`. 

### Major
- Remove the unnecessary step to get the ingress value https://github.com/entando/entando-docs/pull/724#discussion_r1184337791_
- The step "APIs" has to be improved
    -  explicitly explained, this is not an installation step but is used for checking. It could move at the end of the document under an explicit verification chapter
    - Currently, this step doesn't contain enough data to be run as is, we should suggest a way to retrieve the bearer token, then call the API using curl (Postman is also a solution), and maybe provide a zip file to upload. Finally, we should provide a sample command to run it: https://github.com/entando/entando-docs/pull/724#discussion_r1184681393_
- The step "Configure the Entando App Engine" has to be improved
    - The "CDS_PUBLIC_URL" value should come from the ingress defined before, we can provide a way to get it or use the same placeholder as the descriptor file
    - It's not clear what the "CDS_PRIVATE_URL" should contain, is it the service URL?
    
My suggestion:
overall the tutorial is working if we know how to proceed with K8S resources. Reorganizing some steps could help to have a better understanding of what is related to the CDS, what is related to the Entando app configuration, and which step is related to testing the system.

### Bugs
- Using the CDS in the Entando-De-App causes a bug for the app-builder menu
    - The app-builder is trying to get the menu js file using a request but the CDS returns a 404 error because it is empty, the component manager has previously loaded the menu file into the previous de-app volume, not the CDS.

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: YOUR-CDS-DEPLOYMENT
+  template:
+    metadata:
+      labels:
+        app: YOUR-CDS-DEPLOYMENT
+    spec:
+      containers:
```
avdev4j, line 108:
```suggestion
      initContainers:
        - name: init-cds-data
           image: busybox
           imagePullPolicy: IfNotPresent
           command: ['sh', '-c', 'wget --no-check-certificate https://raw.github.com/entando-ps/cds/pr-it-317/entando-data/entando720.tar.gz; tar -xf entando720.tar.gz; rm entando720.tar.gz ']
           volumeMounts:
             - mountPath: /entando-data
                name: cds-data-volume
      containers:
```

nshaw: @avdev4j This doesn't look like the final path since it includes the PR in it - cds/pr-it-317/.  This would be better for now https://github.com/entando-ps/cds/blob/entando720/entando-data/entando720.tar.gz?raw=true but I think we actually need to tag the repo so we can pull a raw file based on the version of Entando.  @pietrangelo what do you suggest here?

vuepress/docs/next/tutorials/consume/mt-cds.md
jyunmitch: it was edited with placeholders

vuepress/docs/next/tutorials/consume/mt-cds.md
```diff
@@ -0,0 +1,285 @@
...
+  selector:
+    matchLabels:
+      app: YOUR-PRIMARY-CDS-DEPLOYMENT
+  template:
+    metadata:
+      labels:
+        app: YOUR-PRIMARY-CDS-DEPLOYMENT
+    spec:
+      initContainers:
+	        - name: init-cds-data
```
avdev4j, line 110: @jyunmitch I guess the suggestion made with the GitHub UI has broke the indentation, we should remove the part in red here (it seems to be equivalent to one tabulation

vuepress/docs/next/tutorials/consume/mt-cds.md
avdev4j: @nshaw Yes, I used the path given by PJ for testing. since we have the final we can update the doc with cc @jyunmitch

vuepress/docs/next/tutorials/consume/mt-cds.md
jyunmitch: edited the quickstart names with placeholders

nshaw:
Replacing this with the new PR https://github.com/entando/entando-docs/pull/724 which I think addresses the ideas and concerns raised by @avdev4j

PR #753, opened 2023-06-27
ENDOC-722 Setup the 6.3.2 patch code logic

nshaw:


nshaw:
Depending on the timing we can either merge and publish this with 7.2.2, or push a new PR later with just the 6.5.4 patch version change.

jyunmitch (approved):

PR #756, opened 2023-07-19
ENDOC-705-update-4Cimage

jyunmitch:


nshaw (approved):
Do we want this change on 7.2 as well?

PR #760, opened 2023-07-24
ENDOC-723-v1-v5

jyunmitch:


vuepress/docs/next/docs/reference/convert-bundle.md
```diff
@@ -55,6 +55,8 @@ PLATFORM_FOLDERS_NAME = [
 'contentModels',
 'resources'
 ]
+* If the v1 bundle used the `ingressPath` specification in a microservice, it should be removed and replaced with an API claim. To create an API, use the `ent bundle api add` command and for more information, see the [API Management page](../getting-started/ent-api.md).
```
jyunmitch, line 58: change this link to ms Add an API Claim tutorial

vuepress/docs/next/docs/reference/convert-bundle.md
```diff
@@ -22,7 +24,42 @@ The improvements include:
...
+'pages',
+'pageTemplates',
+'contentTypes',
+'contentTemplates',
+'contents',
+'assets',
+'groups',
+'labels',
+'languages',
+'pageModels',
```
nshaw, line 54: pageTemplates is the same as the older pageModels, and contentTemplates is the same as contentModels. Maybe the deprecated model names could be added as a note rather than mixed in here whether the duplication might be confusing?

nshaw (approved):

vuepress/docs/next/docs/reference/convert-bundle.md
jyunmitch: @nshaw if the older v1 bundle had a contentModel folder, will it be converted to contentTemplates folder or stay in the same contentModel folder?

PR #771, opened 2023-08-28
ENDOC-743-create-v73

jyunmitch:


vuepress/docs/.vuepress/config.js
```diff
@@ -173,14 +177,15 @@ module.exports = {
...
+        "v73": "v7.3.0",
         "v72": "v7.2.2",
         "v71": "v7.1.6",
         "v70": "v7.0.2",
         "v632": "v6.5.4"
       },
       logoLink: "https://entando.com",
       section: "Docs",
-      version: "7.2",
+      version: "7.3",
```
nshaw, line 188: This change should be left until you're ready to activate 7.3

nshaw (changes requested):
I don't see the change to package.json which we discussed but other than that and the one change in config.json, this looks good to me. The last check I would make, with the package.json tweak, is to run yarn docs:syncNext. If you did it all correctly, you should get no differences with this setup since next and 7.3 should completely the same, other than landing pages which have the version warnings.

nicpuddu (approved):

nshaw (approved):

PR #776, opened 2023-09-07
ENDOC-740-ms-resource

jyunmitch:
Is there anywhere else that needs mention of the v6 descriptorVersion change?


nshaw:
@jyunmitch It would be worth noting in the release notes for sure. Not sure if you want to include that in the PR or track it specifically in a release notes ticket like we did in the past.

nshaw (approved):

jyunmitch:
> @jyunmitch It would be worth noting in the release notes for sure. Not sure if you want to include that in the PR or track it specifically in a release notes ticket like we did in the past.

Yes, I started a release notes list for 7.3 and will add this item

PR #812, opened 2024-01-19
ENDOC-778-plugin-path

jyunmitch:


vuepress/docs/next/docs/curate/bundle-details.md
```diff
@@ -103,15 +103,17 @@ The following is a list of specifications for the bundle descriptor and its comp
...
 |`permissions`|[Permission[]](#permission-specification)|No| | List of permissions to grant to the microservice |
 |`resources`|[resources[]](#microservices-resources)| No | | Requested resources for storage, memory and CPU |
 |`roles`|String[]|No||Exposed security roles|
 |`stack`|Enum|Yes|*spring-boot<br/>*node<br/>*custom|Microservice stack |
 |`version`|String|Required only for a custom stack||Microservice version override|
 
-**^ dbms none**: Oracle and other DBMS types are not supported for automatic deployment in a container. Bundle env variables should be used instead, similar to connecting the EntandoApp to an [external database](../../tutorials/devops/external-db.md).
+**^ `dbms` none**: Oracle and other DBMS types are not supported for automatic deployment in a container. Bundle env variables should be used instead, similar to connecting the EntandoApp to an [external database](../../tutorials/devops/external-db.md).
+
+**^^ `ingressPath`**: If the `ingressPath` is used, note that the `SERVER_SERVLET_CONTEXT_PATH` [environment variable for microservices](#microservices-environment-variables) must also be specified with this custom path.
```
antromeo, line 116: it should be something like:
> If the `ingressPath` is used, ensure that the microservice will accept connections on both `SERVER_SERVLET_CONTEXT_PATH` and `ingressPath`. Entando operator will populate the `SERVER_SERVLET_CONTEXT_PATH` properly and use it for healthChecks, instead `ingressPath` will be used to add an additional route into the ingress.

@firegloves do you think it's clear enough?

firegloves: @antromeo Yes, I do. Do we want to mention that `SERVER_SERVLET_CONTEXT_PATH` is automatically injected by the CM and so it has to be tweaked manually?

jyunmitch: @antromeo @firegloves, according to [ENG-5232](https://entando.myjetbrains.com/youtrack/issue/ENG-5232) Fix plugins canonical and custom ingress path generation, SERVER_SERVLET_CONTEXT_PATH is auto set with the canonical path and so the user has to insert a custom path. 
1. is the custom path = ingressPath? 
2. "Entando operator will populate the SERVER_SERVLET_CONTEXT_PATH properly and use it for healthChecks, instead ingressPath will be used to add an additional route into the ingress." This is confusing since the engr ticket said it is populated by the canonical path and will not use the custom path. If it has to be tweaked, how is it being set properly in cases where a custom path is needed?

jyunmitch: @antromeo, ignore my previous questions. After discussing it briefly with Nicola P., I now understand that the ticket was not solving the issue as I had thought. 

How about this for line 116
If the `ingressPath` is used, note that the `SERVER_SERVLET_CONTEXT_PATH` [environment variable for microservices](#microservices-environment-variables) will also need to be specified manually. The Entando Operator populates the variable automatically and can not take the custom ingress path into account.

in this case, less detail might be better

firegloves: @jyunmitch unfortunately the situation is a bit complex and I can't provide a certain answer before testing the situation and ensuring that there is a way to solve this puzzle. And unfortunately I have no time to test it for some days :(

jyunmitch: @firegloves that's ok, take your time.

antromeo: mmh no, I did some tests with this bundle [romeocontainers/20230502-python-ms](https://hub.docker.com/r/romeocontainers/20230502-python-ms/tags), the tag 0.0.5 has `SERVER_SERVLET_CONTEXT_PATH` specified manually, so you have in entando.json
```
"env":  [{
                "name": "SERVER_SERVLET_CONTEXT_PATH",
                "value": "/pythonmscontextpath"
        }]
```
instead the tag 0.0.4 doesn't have the variable `SERVER_SERVLET_CONTEXT_PATH`.

In both tags, you have in entando.json:
 ```
 "ingressPath": "/pythoncustompath",
 "healthCheckPath": "/api/health"
 ```
 When you install the bundle you get the following cases:
 
 * tag 0.0.4
    -  `SERVER_SERVLET_CONTEXT_PATH` populates automatically with value: `/20230502-python-ms-d3bcdf80-986a1b71/python-ms`
    -  rules in ingress that exposes `/20230502-python-ms-d3bcdf80-986a1b71/python-ms` and `/pythoncustompath`
    so in this case I was suggesting to expose the application in both contexts, so that the application also accepts connection via `ingressPath`.
    
  * tag 0.0.5
    -  `SERVER_SERVLET_CONTEXT_PATH` populates with value: `/pythonmscontextpath`
    -  rules in ingress that exposes `/20230502-python-ms-d3bcdf80-986a1b71/python-ms` and `/pythoncustompath`
  
  In both cases the healthCheckPath in configuration has been appended to liveness, readiness, startup probes, so we have: 
 `/20230502-python-ms-d3bcdf80-986a1b71/python-ms/api/health`
   so in second case the microservice pod will never be ready, because currently `healthCheckPaths` are built using the automatically generated canonical path.
   In my knowledge the fix on `healthCheckPath` has been stopped here [ENG-4990](https://entando.myjetbrains.com/youtrack/issue/ENG-4990/Let-the-user-choose-if-the-canonical-or-custom-ingress-path-must-be-used-for-the-health-check), instead the ticket [ENG-5232](https://entando.myjetbrains.com/youtrack/issue/ENG-5232) was supposed to check that ingress is properly populated when the `ingressPath` is used (an additional rules in ingress), and it it actually done.
   Let me know if it is clear now, I asked @firegloves to speak asap to verify this is the expected behavior

firegloves: We decided to stop the `healthCheckPath` change because we identified the health check as an Entando internal (so it doesn't need to be tweaked by users). About the health check path, the user must be responsible only for the subpath.

Might it be solvable by setting the health check path properly?

However, I'm not sure how much important is this feature, and I'm not sure that spending so much time on this is worth

jyunmitch: @firegloves @antromeo Since there's no resolution, I think it best not to document anything right now and drop this PR. IF you have other suggestions, let me know. Thank you for the support.

firegloves: @jyunmitch unfortunately I agree

